PR #27, opened 2026-02-17
feat: embed shell aliases natively, base+session layout

YardenDagan:
## Summary
- Add `--session` flag and `shell-setup` command for native co-op aliases (`amc`/`amx`)
- Refactor to base+session layout, drop `AM_SESSION` in favor of `AM_ROOT` always pointing to a session subdirectory
- Bump AMQ CLI version to 0.18.0, skill version to 1.2.0

## Test plan
- [ ] `make ci` passes (vet, lint, test, smoke)
- [ ] `amq coop exec claude` sets up correct session layout
- [ ] `amq shell-setup --install` creates working aliases
- [ ] Isolated sessions via `amc feature-a` / `amx feature-a` work correctly


avivsinai (changes requested):
Hey @YardenDagan — first off, thank you for putting this together! Community contributions like this are what make open source great, and it's clear you put real thought and effort into this. The ergonomic problem you identified is spot-on — `amq coop exec --root .agent-mail/feature-x claude` is way too verbose, and both `--session` and shell aliases are features we genuinely want to add.

We ran this through a thorough review (multiple reviewers + Codex via AMQ + external LLM council) and here's the consolidated feedback.

## CI is broken

- `gofmt` fails on `internal/cli/common.go` (struct field alignment in `commonFlags`)
- `golangci-lint` fails with 3 errcheck issues in `shell_setup.go:257` and `shell_setup_test.go:84,97`

## Critical: `--root` semantics silently changed

This is the main blocker. The PR redefines what `--root` means throughout the codebase:

- `defaultRoot()` now returns `.agent-mail/team` instead of `.agent-mail` (`common.go:103-113`)
- `amq env --root /tmp/root` now produces `/tmp/root/team` (`env.go:85-88`)
- `amq coop exec --root .agent-mail/auth` now resolves to `.agent-mail/auth/team` (`coop_exec_unix.go:110-114`)

This breaks every existing script, CI pipeline, and env that uses `AM_ROOT` or `--root`. There's no migration tool and the stderr warning from `warnLegacyLayout()` isn't visible in non-interactive contexts.

**Repro:**
```bash
go run ./cmd/amq env --root /tmp/amq-root/team --json
# Produces: /tmp/amq-root/team/team  (double-session bug)
```

## Recommended approach: `--session` as pure sugar

Instead of changing what `--root` means, add `--session` as a shorthand that composes with the existing root:

```
--session auth  →  --root <resolved_base>/auth
```

That's it. No changes to:
- What `AM_ROOT` means (still the literal queue root)
- What `--root` means (still the literal path)
- What `.amqrc` `root` means (still the literal queue root)
- `defaultRoot()` (still returns `.agent-mail`)

Zero breaking changes. Zero migration needed. Same ergonomic win.

## Shell aliases: eval pattern only

The `amq shell-setup` command generating shell functions is great. But:

- **Remove `--install`** — don't modify user rc files. Use the standard `eval "$(amq shell-setup)"` pattern (like direnv, rbenv, nvm).
- **Remove `promptShellSetup()` from `coop init`** — interactive prompts break scriptability. `coop init` has a `--json` flag, implying it should be scriptable.
- Document the eval pattern in README and let users add it to their rc files themselves.

## Other issues

- **`guardRootOverride`** adds `common.validate()` to 15+ command handlers and errors when `--root` differs from `AM_ROOT`. This is over-protective — a warning would be more appropriate, and with the pure-sugar approach it's not needed at all.
- **`"team"` as default session name** is opinionated and conflicts with swarm mode terminology (Claude Code Agent Teams). With pure sugar, there's no default session name needed.
- **CHANGELOG** is one line for what's actually a multi-faceted change.
- **CLAUDE.md CLI reference** doesn't list `shell-setup`. `amq env` docs don't show `--session`.
- **`TestInstallToRCFileIdempotent`** never actually calls `installToRCFile` — it only tests marker detection.
- **`TestDetectShell`** uses `os.Setenv` instead of `t.Setenv`.

## Suggestion: split into two PRs

These are independent features with different risk profiles:

1. **PR A**: `--session` flag as pure sugar on `coop exec` and `env` (small, safe, high value)
2. **PR B**: `amq shell-setup` command with eval-based aliases (additive, no rc modification)

This makes each PR smaller, easier to review, and independently mergeable.

## Summary

The feature direction is exactly right — we want both `--session` and shell aliases. The implementation just needs to be non-breaking. Keep `--root` meaning what it always meant, add `--session` as sugar on top, and use the eval pattern for shell setup.

Really appreciate you tackling this — happy to help iterate and get this across the line! 🙏

PR #84, opened 2026-04-10
feat: agent retains coop awareness after /clear

omriariav:
## Summary

Fixes #71 — agent loses AMQ awareness after `/clear` or compaction.

- **New CLI command**: `amq integration claude context` reads durable coop state (AM_ROOT, AM_ME, .amqrc, peer presence, inbox) and emits a deterministic preamble for context re-injection
- **SessionStart hook updated**: `claude-session-start.sh` now calls the new command and emits `hookSpecificOutput.additionalContext` + `systemMessage`, so Claude retains session awareness after `/clear`
- **17 unit tests** covering preamble generation, peer discovery, inbox counting, hook JSON output, and help/error paths
- **CLAUDE.md** updated with new command reference

### Design decisions (per Aviv's comment on #71)

- Go CLI is the single source of truth for session/peer discovery — the hook script is a thin shell wrapper
- SessionStart handles baseline identity recovery; UserPromptSubmit stays for dynamic deltas only
- `--event` flag is informational (no behavior change) to support future hook-specific logic

### Preamble output

```
AMQ coop active: me=claude session=collab project=pm-context peers=codex(active).
Inbox: 2 unread message(s). Run: amq drain --me claude
Use amq send/reply/drain for peer coordination. Preserve thread IDs on replies.
```

## Test plan

- [x] `go test ./...` — all 14 packages pass
- [x] `go build ./...` — compiles clean
- [x] Manual QA: plain text output, JSON output, hook script output
- [x] Tested across: solo session, multi-peer session, bare root, no AM_ROOT
- [x] Existing registry tests updated (commandNames, childNames for integration)


omriariav:
## Codex Review Findings (addressed in 8ee22f9)

Ran Codex as a reviewer on the initial commit. Four findings, all addressed:

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| 1 | **High** | `AMQ coop active` emitted without a real session — `resolveRoot(".agent-mail")` can return non-empty even without an active coop | Added `agents/` dir existence check; exits with `ExitNotFound` if missing |
| 2 | **Medium** | Hook phase 2 re-runs `amq integration claude context` without passing AM_ROOT/AM_ME from phase 1 — can diverge or produce bogus context | Phase 1 now captures resolved values and passes them via `--root`/`--me` flags to phase 2 |
| 3 | **Medium** | `--me` not validated through `normalizeHandle` — uppercase/invalid handles silently accepted | Added `normalizeHandle` call with `ExitUsage` on failure |
| 4 | **Low** | `marshalClaudeHookOutput` helper unused in shipped path (hook script builds its own JSON) | Removed unused helper; added edge-case tests: no-agents-dir, invalid handle, JSON output golden |

Test results: `go test ./...` — 14 packages, 0 failures.

omriariav:
@avivsinai — ready for your review. Addresses your approach from #71 (SessionStart hook + Go CLI as single source of truth).

avivsinai (changes requested):
Thanks for picking this up, Omri — #71 is a real paper-cut and SessionStart is the right hook for it. I reviewed this with Codex across two rounds; this comment is our consolidated feedback at HEAD `8ee22f9`. **We're asking for a scope shrink, not a rewrite** — the change you need is much smaller than what's in the PR today, and most of the bugs below evaporate if you take the minimal path.

## TL;DR

The minimum fix for #71 is **~30–60 lines of shell** in `scripts/claude-session-start.sh` (phase 2), composed from existing AMQ primitives. The PR ships ~540 net new lines because `internal/cli/integration_claude.go` re-implements things `amq env --json`, `amq presence list --json`, and `amq list --new --json` already do. If you delete the Go surface and do the minimal shell rewrite, the three blockers below go away for free.

## Blockers (must fix before merge)

### 1. Hook JSON is contract-non-compliant — re-injection may not fire

`scripts/claude-session-start.sh:71-76` emits:

```json
{"hookSpecificOutput":{"additionalContext":"…"},"systemMessage":"…"}
```

The universal schema at https://code.claude.com/docs/en/hooks says verbatim:

> `hookSpecificOutput` is a nested object for events that need richer control. It requires a `hookEventName` field set to the event name.

Without `hookEventName: "SessionStart"`, Claude Code may discard `additionalContext` entirely — which is the exact failure mode #71 is supposed to fix. **This is the single most load-bearing line in the whole PR.**

Required shape:

```json
{
  "hookSpecificOutput": {
    "hookEventName": "SessionStart",
    "additionalContext": "AMQ coop active: me=claude session=collab project=… peers=…"
  },
  "systemMessage": "AMQ coop: 1 peer, 2 unread"
}
```

### 2. `gofmt` fails — `make fmt-check` exits 1

`internal/cli/integration_claude.go:44-50` has struct-tag misalignment and a trailing blank line at EOF. `gofmt -d internal/cli/integration_claude.go` produces a real diff. The pre-push hook in `scripts/install-hooks.sh` would have caught this — worth running `./scripts/install-hooks.sh` locally so future pushes get the full `make ci` gate before they hit CI.

This goes away entirely if the Go file is deleted per the rework below.

### 3. `resolveProject` leaks the caller's cwd project into the preamble

`internal/cli/integration_claude.go:112-113` calls `resolveProject(root)`, which goes through `findAmqrcForRoot` in `internal/cli/env.go`. When `--root` is supplied without an `.amqrc`, the helper falls back to a cwd-based `.amqrc` lookup — so the preamble gets poisoned with an unrelated project name.

Reproducer:

```bash
tmp=$(mktemp -d)
mkdir -p "$tmp/sessionx/agents/claude/inbox"/{new,cur}
# Run from the agent-message-queue repo root:
go run ./cmd/amq integration claude context --root "$tmp/sessionx" --me claude --json
```

Output:

```json
{
  "me": "claude",
  "project": "agent-message-queue",   ← leaked from cwd, not from --root
  "preamble": "AMQ available: me=claude project=agent-message-queue…"
}
```

If a SessionStart hook fires from a directory that isn't the coop session's project, Claude receives a false routing cue. This bug also disappears with the minimal rewrite below — `amq env --me "$ME" --json` resolves `project` correctly because it uses the full resolver chain.

## The scope issue (the real feedback)

Before this PR, `scripts/claude-session-start.sh` was 34 lines and phase 1 already called `amq env --me "$ME"`. The repo already exposes everything the preamble needs:

| Need | Existing primitive |
|---|---|
| root, me, project, session_name | `amq env --me "$ME" --json` |
| peer activity (active/idle) | `amq presence list --json` |
| unread count | `amq list --me "$ME" --new --json` |

`internal/cli/integration_claude.go` re-derives the first two and reinvents the third with weaker error handling (`countNewMessages` returns `0` silently when the directory is unreadable, lying about an empty inbox). We'd rather compose the existing CLI than maintain a parallel pipeline.

Secondary concerns with the new Go surface (any one is a soft "no" on its own; together they're a definitive "delete it"):

- **Wrong semantic home.** `docs/adapter-contract.md` defines `integration` as *"converters from external lifecycle/task events into AMQ messages."* This command converts nothing and emits no messages — it's local context synthesis.
- **Dead CLI flag.** `integration_claude.go:65, 85` declares `--event` and silently discards it (`_ = eventFlag`). Shipping reserved-for-future flags is cruft.
- **Silent failure in `countNewMessages`.** `integration_claude.go:171-184` — `ReadDir` errors become "0 unread", indistinguishable from a genuinely empty inbox.
- **Silent classification in `discoverPeers`.** `integration_claude.go:141-168` — `presence.Read` errors and malformed `LastSeen` timestamps both render as `(stale)`, hiding data corruption.
- **Parallel root-resolution chain.** `integration_claude.go:87` uses `resolveRoot(common.Root)` (a subset of `env.go`'s resolver), so it can return `NotFoundError` for a root that `amq env` would have resolved via `AMQ_GLOBAL_ROOT` or `~/.amqrc`.

## Proposed rework

### Revert

- `internal/cli/integration_claude.go`
- `internal/cli/integration_claude_test.go`
- The `claude` case in `internal/cli/integration_symphony.go` dispatcher
- The `claude` entry in `internal/cli/registry.go` and the corresponding `registry_test.go` line
- The `amq integration claude context` row in `CLAUDE.md`

### Rewrite phase 2 of `scripts/claude-session-start.sh`

Sketch (you'll want to polish — this is to show the shape, not the final text):

```bash
# --- Phase 2: Context re-injection ---
if command -v amq >/dev/null 2>&1; then
    ENV_JSON=$(cd "$PROJECT_DIR" && amq env --me "$ME" --json 2>"$HOOK_LOG") || ENV_JSON=""
    PRESENCE_JSON=$(amq presence list --json 2>"$HOOK_LOG") || PRESENCE_JSON="[]"
    INBOX_JSON=$(amq list --me "$ME" --new --json 2>"$HOOK_LOG") || INBOX_JSON="[]"

    if [ -n "$ENV_JSON" ]; then
        python3 - "$ME" <<PY
import json, sys
me = sys.argv[1]
env = json.loads('''$ENV_JSON''')
presence = json.loads('''$PRESENCE_JSON''')
inbox = json.loads('''$INBOX_JSON''')

# Build peer list — active peers only, from presence
peers = [p for p in presence if p.get("handle") != me and p.get("status") == "active"]
peer_str = ",".join(p["handle"] + "(active)" for p in peers) if peers else ""

session = env.get("session_name", "")
project = env.get("project", "")
unread = len(inbox)

line1 = f"AMQ coop active: me={me}"
if session: line1 += f" session={session}"
if project: line1 += f" project={project}"
if peer_str: line1 += f" peers={peer_str}"
line1 += "."

parts = [line1]
if unread: parts.append(f"Inbox: {unread} unread message(s). Run: amq drain --me {me}")
parts.append("Use amq send/reply/drain for peer coordination. Preserve thread IDs on replies.")

preamble = "\n".join(parts)
banner = f"AMQ coop: {len(peers)} peer(s)"
if unread: banner += f", {unread} unread"

print(json.dumps({
    "hookSpecificOutput": {
        "hookEventName": "SessionStart",   # REQUIRED per universal schema
        "additionalContext": preamble,
    },
    "systemMessage": banner,
}))
PY
    fi
fi
```

Two important details:

1. **Use `amq presence list --json` for peers**, not `env.peers`. `env --json` returns the cross-project `.amqrc` federation map, not the current session's active agents.
2. **Tighten error handling.** Narrow the current `except Exception: pass` to `(json.JSONDecodeError, KeyError, TypeError)` and redirect swallowed stderr to a debug log (e.g. `HOOK_LOG="${TMPDIR:-/tmp}/amq-hook-$USER.log"` at the top of the script). Silent hooks are fine; undiagnosable hooks are not.

### Add a smoke test for the hook script

The repo already has the pattern for shell+python hook coverage:

- `scripts/smoke-test.sh` — wired via `make ci`
- `scripts/test_session_name.py` — unit-tests the notify hook's parallel Python logic

`claude-session-start.sh` has more logic than the notify hooks (two phases, `sed` quote-stripping, JSON extraction) and is a higher-visibility entry point than either of them. Please extend the existing pattern with a fixture that:

- sets `CLAUDE_ENV_FILE` to a temp path
- seeds `AM_ROOT` with a fixture session that has 1 peer + 2 unread messages
- invokes the script
- asserts `$CLAUDE_ENV_FILE` contains `export AM_ROOT=…`
- captures stdout and `json.loads` it, asserts `hookSpecificOutput.hookEventName == "SessionStart"`, `additionalContext` is non-empty, and `systemMessage` mentions the peer count

### What NOT to add

Please keep this PR narrow. Several adjacent improvements are tempting but belong in a separate PR:

- ❌ An `amq install-hooks` / hook installer command
- ❌ `amq doctor --ops` checks for whether the hook is wired
- ❌ Renaming `claude-session-start.sh` to something agent-agnostic
- ❌ Codex CLI hook coverage

We're tracking all of those as a follow-up — see below. They're real gaps, but bundling them here would just recreate the scope problem in a different form.

## Expected shape after rework

- `scripts/claude-session-start.sh` grows by ~30–50 lines (phase 2)
- All the Go changes go away (~485 lines deleted)
- One new smoke-test block or file under `scripts/`
- Three blockers resolved, most of the silent-failure findings resolved
- Net diff: roughly **+60 / –485** instead of the current **+561 / –19**

## Follow-up we're opening separately

Research during this review turned up a broader gap: the hook install UX is a bash echo of commented JSON in `setup-coop.sh:112-127`, there's no `doctor --ops` coverage for hook wiring, and Codex CLI now supports lifecycle hooks too (via `features.codex_hooks = true` + `~/.codex/hooks.json`, with the same `hookSpecificOutput.hookEventName` contract — verified against source at [`codex-rs/hooks/src/events/session_start.rs`](https://github.com/openai/codex/blob/main/codex-rs/hooks/src/events/session_start.rs)). We'll open a separate issue for:

- `amq install-hooks --claude --codex` installer (idempotent JSON/TOML merge, project-local and `--user` scopes)
- `doctor --ops` hook-wiring check for both agents
- Agent-agnostic rename (`amq-session-start.sh`) once the installer can write the new path intentionally
- Same shell script serving both agents (output contract is identical)

None of that blocks this PR.

## What we got right in PR #84

Worth calling out — several things are cleanly done and we want them preserved through the rework:

- Scope discipline — nothing outside #71 snuck in
- The `setupCoopFixture` test helper (clean, DAMP, reusable — keep it if any Go stays)
- `ExitCodeError` assertions verify the actual error kind rather than just presence
- 0700/0600 permissions on test fixtures
- You already iterated once on Codex feedback in `8ee22f9` — the engagement is visible and appreciated
- SessionStart is the right event for #71 (verified against Claude Code docs — fires on `startup|resume|clear|compact` by default when matcher is omitted)

Thanks again for jumping on this. Let us know if anything in the rework plan is unclear or if you'd like to pair on the shell rewrite — happy to help unblock.

omriariav:
@avivsinai — reworked per your review. Here's what changed in `336c8d4`:

## What was done

**Reverted all Go changes** — `integration_claude.go`, test file, dispatcher case, registry entry, CLAUDE.md line. Zero diff from `main` on those files.

**Rewrote phase 2** of `claude-session-start.sh` to compose existing CLI primitives:

| Need | Primitive used |
|------|----------------|
| root, me, project, session | `amq env --me "$ME" --json` |
| peer activity | `amq presence list --json` |
| unread count | `amq list --me "$ME" --new --json` (piped to Python for count) |

Python heredoc assembles the preamble and emits the hook JSON with `hookEventName: "SessionStart"`.

**All three blockers resolved:**
1. ✅ `hookEventName: "SessionStart"` present in `hookSpecificOutput`
2. ✅ `gofmt` failure gone (Go file deleted)
3. ✅ `resolveProject` cwd leak gone (`amq env --json` uses the full resolver chain)

**Additional fixes** (from Codex MCP review of the rework):
- `RESOLVED_ROOT` now captured from existing env file when phase 1 skips the append (`/clear` scenario where `AM_ROOT` is in the env file but not in parent shell env)
- Inbox JSON passed via pipe → only the integer count hits argv (avoids size limits on large inboxes)
- Narrow exception handling: `(json.JSONDecodeError, KeyError, TypeError)` instead of `except Exception`
- Debug stderr from phase 2 redirected to `$HOOK_LOG` (`${TMPDIR}/amq-hook-${USER}.log`)

**Smoke test added** to `scripts/smoke-test.sh`:
- Seeds fixture: 1 peer (codex, active presence) + 2 unread messages
- Asserts phase 1: `export AM_ROOT=…` in env file
- Asserts phase 2: valid JSON with `hookEventName == "SessionStart"`, non-empty `additionalContext`, peer + unread count in `systemMessage`
- Second test case: `/clear` scenario (env file pre-populated, `AM_ROOT` not in parent env)

## Net diff from main

```
 scripts/claude-session-start.sh | +104 -19
 scripts/smoke-test.sh           |  +81
```

All Go files unchanged from `main`. `fmt-check`, `vet`, `go test ./...` (14 packages), and `smoke-test.sh` all pass.

avivsinai (approved):
Thanks Omri — and thanks Codex for the follow-up pass.

Recap of what landed since the last review:

**Reworked at `336c8d4`** (Omri):
- All Go reverted; phase 2 now composes `amq env --json` / `amq presence list --json` / `amq list --new --json` in shell.
- `hookEventName: "SessionStart"` present in hook output.
- `gofmt` blocker gone (Go file deleted).
- `resolveProject` cwd-leak gone.
- Smoke test extended with `/clear` scenario.

**Hardened at `91df083`** (Codex, pushed via maintainer access):
- Fixed bash 3.2 + `set -u` + empty-array regression on stock macOS — replaced `ROOT_FLAGS=()` pattern with explicit rooted/non-rooted command branches.
- Replaced `sed`-based AM_ROOT/AM_ME decoding with isolated `/bin/sh` eval of the matched `export` line — correctly round-trips POSIX single-quote escaping.
- Added symmetric `/clear` recovery for env-file-only `AM_ME`.
- Smoke test extended with quoted-root scenario and `/clear` scenario where only the env file carries `AM_ME=alice`.

`make ci` and `bash scripts/smoke-test.sh` both pass at HEAD. CI green on the fork build (build, Gitleaks, GitGuardian).

`$HOOK_LOG` rotation is the only remaining item — splitting that to a follow-up issue, not a blocker.

Approving and squash-merging.

Issue #95, opened 2026-04-14
amq wake TIOCSTI injection corrupts in-progress user input


omriariav:
## Problem

When the user is actively typing in Claude Code or Codex CLI and `amq wake` fires a notification, the injected text concatenates with the in-progress user input. The injected `\r` then submits the combined string as a single prompt to the agent — the user's intended prompt gets garbled with the AMQ notification text and the merged thing is sent as an instruction.

## Repro

1. Start `amq wake --me claude &` in a Claude Code terminal
2. Start typing a prompt (do not press Enter yet)
3. Have another agent send a message (or wait for one to arrive) → wake's debounce fires notification injection
4. Press Enter on your in-progress prompt

## Actual behavior

The submitted prompt concatenates user input + AMQ notification. Two real examples from my session:

```
I don't see /email-triage command on a fresh conversation on omri-pmAMQ: message from codex - stream4 intro. Drain with: amq drain --include-body — then act on it
```

```
do you think the current project is fiAMQ
  [stream4]: message from codex - Re: stream4
  intro. Drain with: amq drain --include-body
  — then act on itt
```

Claude Code then tries to interpret the garbled string as a user request.

## Root cause

`internal/cli/wake.go:263` → `injectNotification()` uses TIOCSTI to push notification text + `\r` into the tty's input queue. TIOCSTI is oblivious to readline / Ink / crossterm buffer state in the foreground app — whatever the user has typed-but-not-submitted stays in the app's input buffer, and the injected bytes append to it. The `\r` then commits the whole buffer.

Both `raw` and `paste` modes exhibit this (the "auto" default resolves to `raw` for `claude`/`codex` handles per `effectiveInjectMode()` at wake.go:246).

The existing `--interrupt` path has the same shape (TIOCSTI Ctrl-C + notice) — Ctrl-C alone is benign, but the interrupt notice that follows inherits the same corruption problem if the user is mid-input.

## Impact

- **High** — makes `amq wake` practically unusable while actively composing a prompt
- Data risk: the AMQ notification string becomes a user instruction ("Drain with: amq drain …") and the agent may act on it
- The `\r` submits the garbled prompt, so the user has no chance to correct it

## Why TIOCSTI is the wrong primitive for notifications

Notifications should be out-of-band (terminal emulator surface). TIOCSTI is in-band — it shares the input queue with user typing. Options that stay out-of-band:

| Mechanism | Pros | Cons |
|-----------|------|------|
[...]

PR #97, opened 2026-04-24
fix: defer wake injection during terminal activity

avivsinai:
## Summary

This is an experimental #95 mitigation that preserves the TIOCSTI wake path and adds a bounded input-activity gate before non-interrupt injection.

- Keeps existing wake notification text, inject modes, and CR submission behavior.
- Adds default-on `--defer-while-input` with `--input-quiet-for`, `--input-poll-interval`, and `--input-max-hold` controls.
- Samples the controlling terminal only while a wake injection is already pending: `TIOCINQ`/`FIONREAD` for unread kernel-buffered input plus `fstat` atime for recent foreground reads.
- Fails open if `/dev/tty` or sampling is unavailable, so wake behaves as before rather than wedging.
- Bypasses the gate for explicit urgent interrupt wake messages.

## Performance Shape

Idle path is unchanged: no new polling while there are no pending AMQ messages. During a deferred pending wake, the loop performs one terminal input-queue ioctl and one `fstat` per poll interval, bounded by `--input-max-hold` (default 15s).

## Known Limit

This is a heuristic, not a prompt-buffer guarantee. If Claude Code/Codex already consumed a partially typed prompt into its own editor buffer and the user pauses longer than `--input-quiet-for`, wake can still inject and submit. A true prompt-empty signal would need cooperation from the foreground app.

## Verification

- `go test ./internal/cli`
- `make test`
- `make build`
- `GOOS=linux GOARCH=amd64 go test -c ./internal/cli -o /tmp/amq-cli-linux.test`
- `GOOS=windows GOARCH=amd64 go test -c ./internal/cli -o /tmp/amq-cli-windows.test.exe`
- `git diff --check`
- `./amq wake --help`

## Notes

I intentionally did not rescan the mailbox during the hold loop for priority escalation; that can be a follow-up if testing shows it is needed. This keeps the hold path limited to terminal syscalls and avoids extra mailbox parsing while the terminal is active.


omriariav:
I gave this a local try on macOS/iTerm2/zsh. The automated checks passed for me: `go test ./internal/cli`, `make test`, `make build`, `git diff --check`, and `./amq wake --help`.

For a rough manual test, I ran wake with `--input-quiet-for 60s`, typed text at the prompt, and sent a message after 3s. In that setup wake still injected immediately, so I may not be reproducing the intended Claude/Codex path correctly, but that was the behavior I saw.

One small cleanup note: I think `specs/003-wake-input-deferral/spec.md` may need to be removed before merge per the repo docs policy.

avivsinai:
Pushed `0b55040` to address Omri's manual-test failure and the spec-cleanup nit.

## What changed

- `internal/cli/wake_tiocsti_unix.go`: split `sampleTTYInputState(queueFD, atimeFD)`. `/dev/tty` still drives `TIOCINQ`/`FIONREAD` and `TIOCSTI` injection unchanged. Atime sampling now uses `os.Stdin.Fd()` when stdin is a TTY (the common `amq wake --me claude &` case), falling back to the `/dev/tty` fd otherwise. Added an `atime_source=` debug log.
- `internal/cli/wake_unix.go` + `CHANGELOG.md`: documented Linux's `tty_update_time()` ~8s atime granularity so the heuristic's limits are honest.
- `specs/003-wake-input-deferral/spec.md`: deleted per repo docs policy (spec docs do not live in PR branches).

## Why the path-resolution fix doesn't work

Codex independently checked on macOS: `fcntl(F_GETPATH)` and `ttyname()` on a fd opened from `/dev/tty` both return `"/dev/tty"`, not the underlying `/dev/ttysNNN`. Path-resolution-then-stat is a dead end on Darwin.

## ChatGPT Pro review (gpt-5-3-pro via yoetz)

Consulted ChatGPT Pro on the diagnosis and minimal fix. Operative paragraphs:

> macOS: No, not as implemented. `/dev/tty` on Darwin is an indirect controlling-tty driver. XNU's `tty_tty.c` labels it "Indirect driver for controlling tty"; `cttyread()` resolves the session's controlling tty vnode via `cttyvp(current_proc())` and forwards the actual read with `VNOP_READ(ttyvp, ...)`. That explains the failure mode: read/write/ioctl operations can be redirected to the real `/dev/ttysNNN`, while `fstat(open("/dev/tty"))` still reports status for the open `/dev/tty` descriptor object.

> Recommended fix: prefer `fstat(os.Stdin.Fd())` for last-read time. Use `/dev/tty` fd for TIOCINQ/FIONREAD, eventual TIOCSTI injection, and existing fail-open behavior. `os.Stdin.Fd()` for `fstat()` atime, when stdin is a TTY. This is the smallest reliable improvement because a backgrounded job launched as `amq wake --me claude &` normally inherits fd 0 from the terminal. That fd was open before the user typed and before the foreground app consumed input. On macOS, your repro says this is the fd whose atime tracks the underlying `/dev/ttysNNN`. On Linux, it is also the fd most likely to be present in the tty's open-file list when `tty_update_time()` runs.

> Mostly yes [the branch preserves generic TIOCSTI push capability]. The core push path is preserved: `tiocsti.Inject()` still opens `/dev/tty` and injects with TIOCSTI. Raw mode still injects text, waits, then injects CR twice. Paste mode still uses bracketed paste and delayed CR. Interrupt mode bypasses input deferral because `injectNotification(cfg, interruptText, false)` passes `deferForInput=false`.

> Linux's tty atime update granularity is ~8 seconds, so a 1.2s "quiet" window is more precise than the Linux signal can guarantee. That is a documentation concern even after the fd fix.

ChatGPT's verdict: "approve the design after a small fd-source patch; do not merge the current `/dev/tty`-atime-only implementation." Patch shape they recommended matches `0b55040` directly.

## Verification

Local on macOS arm64:
- `make ci` (fmt-check, vet, lint, test, smoke) — green
- `git diff --check` — clean
- `GOOS=linux GOARCH=amd64 go build ./...` — clean
- `GOOS=windows GOARCH=amd64 go test -c ./internal/cli` — clean

Pure-logic tests in `wake_test.go` (`ttyInputState.active`, `inputDeferralDelay`, `shouldDeferBeforeInject`) still cover deferral behavior. Did not add a fd-selection test — would require either a `term.IsTerminal` seam or a real PTY harness, both heavier than the change warrants.

## Manual reverification needed

@omriariav — the macOS case you tested previously should now defer correctly. Could you re-run with the same setup once CI is green:
```
amq wake --me claude --debug --input-quiet-for 60s &
# type at prompt, send AMQ message ~3s later
```
With `--debug` you should see `atime_source=stdin` and `deferring injection for ... (recent terminal input, pending_bytes=0)` instead of immediate injection.

omriariav:
Manual re-test passed on macOS.

I ran `0b55040` with wake in one pane and sent a message from another pane while typing at the prompt. The previous failure mode was immediate injection into the typed prompt. This time debug showed:

- `input deferral atime_source=stdin`
- repeated `deferring injection ... (recent terminal input, pending_bytes=0)`

It eventually injected only after `--input-max-hold 15s` was reached, which matches the bounded-deferral behavior. This addresses the manual-test failure I saw earlier.

PR #151, opened 2026-06-21
fix: self-heal stale wake locks via verified ownership

ohade:
Minimized stale wake-lock bugfix. Wake repair was split out to draft follow-up #154 so this PR only carries the proven stale-lock self-heal and doctor visibility/fix path.

## Scope

- Make `.wake.lock` an identity-verified ownership record instead of a bare PID marker.
- Store process start token, boot ID, executable, argv, root, agent, and hostname diagnostics.
- Classify locks as stale only when proven: dead PID, root/agent mismatch, non-`amq` process, boot mismatch, or process-start mismatch.
- Treat inconclusive inspection as `unverified`: refuse, never auto-delete.
- Never signal a lock PID unless it is identity-confirmed as the matching `amq wake`.
- Re-read and byte-compare the lock before removal to avoid TOCTOU cleanup.
- Keep legacy locks without new metadata from being auto-deleted merely for missing fields.
- Report stale wake locks from `amq doctor --ops`; `--fix-wake-locks` removes only locks that re-inspect as stale.
- Preserve stale-lock visibility/fix even when `meta/config.json` is missing or corrupt by enumerating `agents/*` for wake-lock checks.
- Add a `CHANGELOG.md` Unreleased entry.

## Split Out

Moved to draft #154:
- persisted `.wake.target`
- `amq wake repair`
- saved-injector execution / detached repair process
- repair availability reporting
- `coop exec --wake-inject-via` / `--wake-inject-arg`

## Testing

- `go test ./internal/cli -run 'TestRunOpsChecks_(NoConfig|ReportsAndFixesStaleWakeLockWithoutConfig|ReportsStaleWakeLock|FixesStaleWakeLock)' -count=1`
- `make ci`

Both are green locally after minimization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Document operator-gate conventions in the `amq-cli` and `amq-spec` skills,
   covering structural human handoffs, initialized human handles, and spec
   approval gates (#136).
+- Report and optionally fix identity-verified stale `.wake.lock` files from
+  `amq doctor --ops`, including roots whose config is missing or corrupt (#151).
 
 ## [0.36.0] - 2026-06-13
 ### Changed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -332,6 +332,7 @@ Use `--force` with retry to override the max retry limit.
 - Queue depth and oldest unread per agent
 - DLQ count and oldest age
 - Presence freshness
+- Wake lock health (`--fix-wake-locks` removes stale locks)
 - Integration hints for Kanban and Symphony
 
 Use `amq doctor --ops --json` for machine-readable health output.
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ amq reply --id <msg_id> --kind review_response --body "LGTM with comments"
 amq doctor
 amq doctor --ops
 amq doctor --ops --json
+amq doctor --ops --fix-wake-locks
 ```
 
 ## Message Kinds & Priority
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -184,9 +184,10 @@ func runDoctor(args []string) error {
 				continue
 			}
 			icon := statusIcons["warn"]
-			if wl.Status == "fixed" {
+			switch wl.Status {
+			case "fixed":
 				icon = statusIcons["ok"]
-			} else if wl.Status == "error" {
+			case "error":
 				icon = statusIcons["error"]
 			}
 			line := fmt.Sprintf("  %s wake lock: %s agent=%s root=%s lock=%s",
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -32,6 +32,7 @@ func runDoctor(args []string) error {
 	fs := flag.NewFlagSet("doctor", flag.ContinueOnError)
 	jsonFlag := fs.Bool("json", false, "Output as JSON")
 	opsFlag := fs.Bool("ops", false, "Include runtime operational checks")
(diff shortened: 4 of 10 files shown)


avivsinai:
Thanks for driving this, Ohad — this is exactly the identity-verified ownership record the stale-lock bug needed, and the test coverage (PID reuse / non-amq / boot-id mismatch / legacy no-auto-delete / reverify-before-signal) is thorough. CI is now approved and running.

Two things before this can land, plus one scope request:

**1. Scope — please split.** This PR does two separable things:
- (a) the proven bug fix: identity-verified stale-lock self-heal + `doctor --ops --fix-wake-locks`;
- (b) a new feature: `amq wake repair` (`.wake.target` persistence + digest binding, permission validation, `setsid` detachment, `coop exec --wake-inject-via/--wake-inject-arg`).

Let's land (a) as this PR — it's the regression Ohad/others actually hit — and move (b) `wake repair` to a follow-up PR. (b) executes a persisted injector and roughly doubles the diff (~+1000 lines), so it deserves its own focused review rather than riding the bug fix. Smallest-composable-change keeps the security-sensitive surface reviewable.

**2. Lint fails.** `golangci-lint run` locally (the gate CI enforces) flags 3 — all trivial:
- errcheck `output.Close()` — `internal/cli/wake_unix.go:424` and `internal/cli/wake_unix_test.go:693`
- staticcheck QF1003 (use tagged switch) — `internal/cli/doctor.go:187`

Build, `go vet`, gofmt, and full `go test ./...` are green locally — only lint is red.

**3. CHANGELOG.** Add a `## [Unreleased]` entry; `release.sh` refuses an empty Unreleased section.

Codex is doing a deeper correctness/security pass on the identity classifier, signal-safety (re-verify before SIGTERM/SIGKILL), the TOCTOU re-read, and the platform inspectors' fail-closed behavior — findings to follow. #153 will be reviewed on top once this lands and rebases onto `main`.

avivsinai:
One more for the minimal (split) PR — found in codex's deeper pass and reproduced:

**`doctor --ops` still misses stale locks when config is missing/corrupt** — which is the exact state Ohad originally reported ("doctor only reported missing config, did not report the stale locks").

`internal/cli/doctor_ops.go:72-80` returns immediately on a `config.LoadConfig` error (emits only `config_error`), and `checkWakeLocks` runs later at `doctor_ops.go:137` iterating `cfg.Agents`. So a root with `agents/codex/.wake.lock` but no `meta/config.json` yields only `config_error` — no `wake_locks`, and `--fix-wake-locks` can't help either. Reproduced on this branch.

Fix: enumerate agents from `agents/*` as a fallback/union for the wake-lock checks even when config load fails, and add a regression test for config-missing + stale lock. Please keep this in the minimal PR — it's core to the doctor-visibility half of the fix, not the `wake repair` feature.

avivsinai:
ChatGPT-Pro owner-lens gate confirms this PR remains blocked in its current shape.

Required before merge:
- drop/split `wake repair`, `.wake.target`, saved-injector execution, repair command/tests/docs, and `coop exec --wake-inject-via/--wake-inject-arg` from this PR;
- keep the minimal identity-verified stale-lock self-heal plus `doctor --ops --fix-wake-locks` visibility;
- add config-missing/corrupt fallback by enumerating `agents/*` as fallback/union for wake-lock checks when config load fails;
- add regression coverage for config-missing + stale wake lock visibility/fix;
- fix the lint issues already listed above;
- add a `CHANGELOG.md` Unreleased entry;
- rerun CI on the minimized diff.

The important owner concern is that persisted `.wake.target` + later executing a saved injector is durable local execution surface, not part of the stale-lock bug fix.

avivsinai (approved):
Approved after final gate chain: Claude review cleared, Codex minimal review found and fixed the pre-signal recheck race in b84330a, ChatGPT-Pro/Yoetz approved the final diff, CI is green, and Ohad's delegated/courtesy gate is satisfied.

PR #153, opened 2026-06-22
fix: coop exec reuses only usable existing wake

ohade:
## Summary

`amq coop exec --require-wake` can now reuse an already-running wake only when the lock is both identity-verified and usable for injection. The transient wake process writes readiness and exits when the existing wake is safe to use; otherwise it fails closed and the agent command is not launched.

This is now rebased onto merged `main` after #151, so the PR contains a single functional change on top of the stale-lock self-heal.

## Behavior

- `coop exec --require-wake` passes internal `amq wake --ready-file ... --accept-existing-wake`.
- `amq wake` writes the ready file for an existing wake only when `--ready-file` and `--accept-existing-wake` are both present.
- The accept path requires the shared wake-lock classifier to report a confirmed valid wake.
- The accept path also checks the existing wake is usable: missing TTY and same-TTY/different-session orphan candidates fail closed instead of reporting readiness.
- Stale, creating, and unverified locks still fail closed under the accept path.
- Normal duplicate `amq wake` behavior is unchanged.
- `waitForWakeReady` accepts the ready file if the transient child writes it and exits immediately.

## Tests

Added/updated coverage for:

- existing usable wake writes the ready file;
- ready-file without `--accept-existing-wake` still rejects an existing wake;
- missing-TTY existing wake rejects and does not write the ready file;
- same-TTY/different-session existing wake rejects and does not write the ready file;
- unverified existing wake remains fail-closed;
- ready file written just before child exit is accepted;
- `Started amq wake` vs `Using existing amq wake` status text.

Verified locally: `go test ./internal/cli ...` focused set and `make ci`.

---
Co-op: authored by Ohad, completed with Codex review/fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Report and optionally fix identity-verified stale `.wake.lock` files from
   `amq doctor --ops`, including roots whose config is missing or corrupt (#151).
 
+### Fixed
+
+- `amq coop exec --require-wake` can reuse an existing usable wake process, while
+  still failing closed when the existing wake cannot safely inject (#153).
+
 ## [0.36.0] - 2026-06-13
 ### Changed
 
--- a/internal/cli/coop_exec_test.go
+++ b/internal/cli/coop_exec_test.go
@@ -135,6 +135,64 @@ func TestCoopExecSessionInvalidName(t *testing.T) {
 	}
 }
 
+func TestWakeReadyMessageReportsExistingWake(t *testing.T) {
+	const wakePID = 4242
+	stubInspectWakeProcess(t, func(pid int) wakeProcessInfo {
+		if pid == wakePID {
+			return wakeProcessInfo{
+				PID:        pid,
+				Running:    true,
+				StartToken: "start-1",
+				BootID:     "boot-1",
+				Executable: "/opt/homebrew/bin/amq",
+				Args:       []string{"/opt/homebrew/bin/amq", "wake", "--me", "codex"},
+			}
+		}
+		return wakeProcessInfo{PID: pid}
+	})
+	root := t.TempDir()
+	writeWakeLockForTest(t, root, "codex", wakeLock{
+		PID:          wakePID,
+		ProcessStart: "start-1",
+		BootID:       "boot-1",
+		Executable:   "/opt/homebrew/bin/amq",
+	})
+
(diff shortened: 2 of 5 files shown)


avivsinai:
Holding this one — correctness concern from the review pass, verified against the code.

**`--accept-existing-wake` can report readiness for a wake that cannot inject.** The accept path returns `wakeLockAlreadyRunningError` (→ ready-file success) for any classifier-`valid` lock *before* `shouldReplaceOrphanedWakeLock` runs (`internal/cli/wake_unix.go:81-84`). But the #151 orphan check treats a confirmed `amq wake` whose `/dev/...` TTY has disappeared as replaceable — i.e. unusable as an injector (`wake_unix.go:167-178`). So under accept, a wake that normal acquire would replace is instead accepted, and `coop exec --require-wake` prints "Using existing amq wake" / signals ready while notifications still can't be delivered. The new `TestRunWakeWithLoopAcceptExistingWakeDoesNotReplaceOrphanCandidate` (TTY=`/dev/amq-missing-tty`) cements that behavior.

Fix options:
- accept only *after* the orphan check confirms no replacement is needed, or
- add a distinct `isWakeUsableForRequireWake` gate and stay fail-closed for the missing-TTY and same-TTY-different-session cases.

Process: this PR is stacked on #151, which we're asking to split (minimal stale-lock + doctor fix first, `wake repair` separately). Once that lands, please rebase this onto it so the review is just commit `1a153c3`. CI is currently red here on the inherited lint issues + missing CHANGELOG.

avivsinai:
ChatGPT-Pro owner-lens gate confirms this PR remains blocked until the reuse path proves wake usability, not just process identity.

Required before merge:
- make `--accept-existing-wake` accept only after the orphan/usability check confirms the existing wake can actually inject;
- missing-TTY and same-TTY-different-session cases must fail closed, not write ready / report "Using existing amq wake";
- flip the missing-TTY acceptance test so it rejects and does not write the ready file;
- keep stale / creating / unverified locks fail-closed;
- rebase this onto the minimized #151 after #151 lands, so review is focused on the coop reuse commit;
- rerun CI.

Identity-confirmed `amq wake` is necessary but not sufficient for `--require-wake`; the existing wake also has to be usable as an injector.

avivsinai (approved):
Approved after final gate chain on a0b07a3: Codex minimal review passed, ChatGPT-Pro/Yoetz approved the corrected final diff, Claude review cleared, CI/Gitleaks/GitGuardian are green, and Ohad courtesy/delegation gate is satisfied.

PR #164, opened 2026-06-22
Fix release workflow merge commits

### LINKED ISSUE #163: Make release workflow robust to merge commits
## Problem

`release.yml` currently only runs the release jobs on a `main` push when `github.event.head_commit.message` starts with `chore(release): v`. If a release PR is merged with a merge commit, the head commit subject is `Merge pull request ...`, so the Release workflow skips even though the release commit and metadata are already on `main`.

This happened during the v0.37.1 recovery: #161 merged the correct release files, but Release skipped; #162 was an empty squash commit with subject `chore(release): v0.37.1` to trigger the workflow without manual tags.

## Desired outcome

Release publication should not depend on humans remembering to squash-merge release PRs. Keep the no-manual-tags rule, but make the workflow detect valid release metadata on `main` safely.

Possible approaches to evaluate:
- Teach `release.yml` to detect a release commit in a merge commit parent / PR payload and use that release version.
- Trigger publication from a protected tag path created by CI only.
- Add an explicit release PR label or metadata gate that works for merge commits as well as squash commits.

## Acceptance

- A release PR merged as a normal merge commit still runs prepare/verify/release/skill-publish for the intended version.
- The workflow still refuses mismatched existing tags and never requires a locally/manual-created tag.
- Existing squash-merge release flow remains supported.

avivsinai:
## Summary
- Move release metadata resolution into `scripts/resolve-release-metadata.sh` so the behavior is unit-testable.
- Let `prepare` run for GitHub merge commits, then resolve the first non-merge `chore(release): v...` subject from the merge side.
- Keep squash release commits supported, keep CI tag creation after verification, and make non-release merge commits exit `prepare` with no outputs so `verify`, `release`, and `skill-publish` skip.

Closes #163.

## Release-engineering checks
- Uses GitHub Actions job outputs through `$GITHUB_OUTPUT` / `needs.prepare.outputs.*`, matching GitHub's documented job-output pattern.
- Keeps `actions/checkout` at `fetch-depth: 0`, which checkout documents as needed for full branch/tag history; the merge-parent scan does not rely on shallow history.
- Uses `::notice::` for the intentional no-op path and `::error::` for refusal paths, matching GitHub workflow command conventions.
- Avoids `git log --grep` for merge detection because it searches commit bodies; the resolver filters subject lines only and excludes merge commits.

Docs checked:
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands
- https://github.com/actions/checkout

## Verification
- `bash scripts/test_release_metadata.sh`
- live history probe: real release merge `139e648` resolves `chore(release): v0.37.1`; non-release path stays empty
- `git diff --check`
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/release.yml`
- `make ci`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,12 @@ concurrency:
 
 jobs:
   prepare:
-    if: "${{ github.event_name == 'workflow_dispatch' || startsWith(github.event.head_commit.message, 'chore(release): v') }}"
+    if: >-
+      ${{
+        github.event_name == 'workflow_dispatch' ||
+        startsWith(github.event.head_commit.message, 'chore(release): v') ||
+        startsWith(github.event.head_commit.message, 'Merge pull request')
+      }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:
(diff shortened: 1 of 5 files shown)


avivsinai:
External review gates passed before merge:

- Claude review: PASS via AMQ on 2026-06-22. Verified merge admission, resolver behavior, empty-tag downstream guards, untouched CI tag creation block, and smoke-test coverage.
- ChatGPT-Pro gate: PASS via Yoetz/ChatGPT Extended Pro. No blocking findings. It confirmed the release gate invariants, merge-commit release SHA behavior, fetch-depth/history prerequisite, and false-positive non-release merge handling.

Local verification run on commit 48be505:
- bash scripts/test_release_metadata.sh
- live history probe: 139e648 resolves v0.37.1; non-release path stays empty
- git diff --check
- actionlint on release.yml
- make ci

Issue #139, opened 2026-06-05
Make the human operator mailbox (`user`) first-class in core


avivsinai:
## Summary

Make the **human operator mailbox** a first-class concept in AMQ core, so the `to:user` "operator gate" pattern proposed in #136 is reliable rather than convention-by-hope.

#136 standardizes the *producer* convention (address the human structurally via `amq send --to user …` instead of burying "operator-held" in agent-to-agent prose) in the **skills/docs** layer. This issue tracks the separate, **core** product decisions that make that signal load-bearing. These are intentionally split so the docs-first PR in #136 isn't blocked on core changes.

## Motivation

Today `user` is just-another-handle:

- It is **not** seeded by `amq init` or `amq coop init`. Against the default `claude,codex` config, `amq send --to user …` warns (`unknown handles [user]`) and **fails under `--strict`**.
- It has no representation in health/observability tooling, so a downstream consumer (e.g. amq-noc) keying `needs-you` off `agents/user/inbox/` has no first-party way to surface "N humans waiting, oldest gate age".
- If a `user` mailbox exists, `who`/presence would render it as a **perpetually-stale agent** (no human process ever updates presence), which is misleading.

## Decisions to make

1. **Seeding** — should `init` / `coop init` seed `user` by default, or behind a flag (e.g. `--with-user` / config)? Default-on removes the warn/`--strict` rough edge but adds a mailbox every project may not use.
2. **`doctor --ops` reporting** — add operator-gate visibility: count of open `to:user` items and oldest gate age, alongside the existing per-agent queue-depth/oldest-unread checks. (No new gate *state* in core unless we decide to model it — see #4.)
3. **`who` / presence** — represent `user` as a human mailbox, **not** an agent process: suppress or distinguish the "stale" liveness semantics so a never-updating `user` doesn't read as a dead agent.
4. **Gate state (open question, likely out of scope for v1)** — decide whether AMQ models explicit gate open/cleared state, or whether clearing stays a pure consumer/orchestrator convention (a reply to the gate thread). Default lean: keep core dumb; routing/classification must **not** special-case `gate/<topic>` threads or `APPROVAL:`/`DONE:` subject text.

## Non-goals / guardrails

- This does **not** make `to:user` an authorization or security boundary. AMQ sender identity is local convention, not authenticated approval. The gate is an observability/handoff signal only.
[...]

PR #165, opened 2026-06-22
Reserve the human operator user handle

avivsinai:
## Summary
- Reserve `user` as the conventional human operator handle for configured roots, without changing no-config validation bypass behavior.
- Update `amq coop init` defaults to seed `claude,codex,user`; leave explicit `amq init --agents ...` unchanged.
- Refresh the operator-gate skill docs so they no longer claim `--to user` fails under `--strict` for configured projects.

Closes part of #139. This is PR1 only: no `who`/presence human rendering, no `doctor --ops` operator gate block, and no gate-state/thread/subject special-casing.

## Landmine Coverage
- `loadKnownAgents` returns `nil` when config is absent and does not synthesize `user` for `agents: []`.
- `loadKnownAgentSet` keeps the nil known-set path for no config.
- Strict header validation accepts `To: ["user"]` when the configured agents are `claude,codex`, but still rejects unknown recipients.
- Strict drain of a valid `claude -> user` message moves to `cur`, not DLQ, and writes a drained receipt.
- Persisted config tests use `config.LoadConfig`, not `loadKnownAgents`, so synthetic validation handles are not confused with stored config.

## Verification
- RED before implementation: focused new tests failed for strict `user` validation, strict drain, and coop defaults.
- `go test ./internal/cli`
- `make check-skills`
- `git diff --check`
- `make ci`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The human operator handle `user` is now reserved for configured projects, and
+  `amq coop init` seeds `claude,codex,user` by default so strict operator gates
+  no longer require custom coop setup (#139).
 - Release publishing now detects release commits inside normally merged release
   PRs while ordinary feature merge commits no-op before tag, artifact, or skill
   publishing jobs (#163).
--- a/internal/cli/common.go
+++ b/internal/cli/common.go
@@ -22,6 +22,8 @@ type commonFlags struct {
 	flagSet *flag.FlagSet
 }
 
+const reservedHumanHandle = "user"
+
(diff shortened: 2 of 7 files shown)


avivsinai:
External review gates passed before merge:

- Claude review: PASS via AMQ on 2026-06-22. Verified the len(cfg.Agents)>0 landmine, no-config/empty-config preservation, strict validator/drain behavior, coop default, explicit init behavior, and PR1 scope split.
- ChatGPT-Pro gate: PASS via Yoetz/ChatGPT Extended Pro. No blocking findings. It confirmed config-backed reservation, no-config and agents:[] preservation, strict header/drain path, coop default seeding, explicit init non-injection, and the skill doc behavior split.

Local verification run on commit 67b5925:
- focused RED run failed before implementation for strict user validation, strict drain, and coop default
- go test ./internal/cli
- make check-skills
- git diff --check
- make ci

PR #167, opened 2026-06-22
Add guarded env export for session shells

avivsinai:
## Summary
- add `amq env --export` as an opt-in shell-output mode for pinning a terminal to the resolved AMQ root
- emit `AM_BASE_ROOT` only when the resolved root is a confirmed session root, including explicit `--session` roots backed by custom/global base roots
- print a stderr pin note while keeping stdout eval-safe, and preserve non-export `amq env` stdout behavior
- document the one-terminal/one-session guardrail in README and the AMQ skill

## Scope
- No change to bare/default `amq env` shell output.
- No `AMQ_GLOBAL_ROOT` export.
- No broader root-resolution precedence changes.

## Verification
- RED before implementation: focused `--export` tests failed because the flag did not exist.
- `go test ./internal/cli -run 'TestRunEnv'`
- `git diff --check`
- `make check-skills`
- `make ci`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `amq env --export` now prints eval-safe shell exports for opt-in terminal
+  pinning, including `AM_BASE_ROOT` only when the resolved root is a session
+  root (#149).
+
 ### Fixed
 
 - `amq who`, `amq presence list`, and `amq doctor --ops` now present the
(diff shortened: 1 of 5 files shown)


avivsinai:
External review gate via AMQ Claude: PASS.

Summary from reviewer:
- `--export` is opt-in; mutually exclusive with `--json` and `--session-name`.
- `AM_BASE_ROOT` is emitted only for session roots; explicit `--session` is authoritative; bare/non-session roots omit it.
- Pin note goes to stderr, keeping stdout eval-safe.
- Non-export output is preserved.
- ChatGPT-Pro not required for this additive guarded change.

Verification:
- GitHub CI/checks green
- local `go test ./internal/cli -run 'TestRunEnv'`, `git diff --check`, `make check-skills`, and `make ci` passed before push

PR #171, opened 2026-06-28
[codex] Add Ghostty wake targets

ohade:
## Summary

Adds native macOS Ghostty wake delivery that targets a stable Ghostty terminal id instead of relying on terminal titles or session labels.

- adds `amq wake --inject-ghostty` and `--inject-ghostty-terminal-id`
- adds `amq coop exec --wake-inject-ghostty` for normal launcher use
- persists `.wake.target` metadata beside `.wake.lock` so wake reuse is target-aware
- reports stale/orphaned/mismatched wake target state in `amq doctor --ops`
- adds `amq doctor --ops --fix-wake-targets` for orphan target cleanup

## Why

In multi-terminal setups, terminal titles can be duplicated or renamed. AMQ routing should remain root/agent based, but terminal injection needs a stable terminal destination. Binding the wake process to a Ghostty terminal id prevents silent injection into the wrong terminal when labels collide.

## Validation

- `go test ./...`
- `git diff --check`
- CLI help smoke checks for `--inject-ghostty`, `--wake-inject-ghostty`, and `--fix-wake-targets`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Report and optionally fix identity-verified stale `.wake.lock` files from
   `amq doctor --ops`, including roots whose config is missing or corrupt (#151).
 
+### Fixed
+
+- `amq coop exec --require-wake` can reuse an existing usable wake process, while
+  still failing closed when the existing wake cannot safely inject (#153).
+
 ## [0.36.0] - 2026-06-13
 ### Changed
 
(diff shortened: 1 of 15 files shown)


avivsinai:
Thanks for the work here. Since this PR was opened, we landed #177, which fixed the stale-owner/reopen regression this was targeting, and the follow-up wake hardening rewrote the wake/doctor/target internals this branch touches. The PR is now draft, stale, and deeply conflicting, so rebasing it as-is would mostly mean redoing those internals against a different design.

The piece we still want is the native Ghostty terminal-id discovery/probe idea. We'd welcome that as a fresh, narrow PR against current main, scoped to the Ghostty terminal-id helper and minimal wiring/docs — ideally integrating through the current `--inject-via` / `coop exec --wake-inject-via` path (a Ghostty bridge example), or a small proposal for a native Ghostty transport if you want that surfaced directly. Please avoid the wake lock/doctor/repair rewrites in this branch.

Closing as superseded/stale to keep the queue clear — this is not a rejection of the Ghostty feature. Happy to help scope the smaller follow-up.

PR #204, opened 2026-07-06
fix: pin session roots to absolute paths (+ reply --wait-for, who base-root header)

ohade:
## Summary

Yesterday I hit a full day of silent message loss between a Claude and a Codex session that were both on "the same" session name. Root cause: `coop exec --session X` resolves to a **relative** `AM_ROOT=.agent-mail/X`, so the physical mailbox re-resolves against each agent's cwd. Two agents in different git worktrees ended up on two disjoint trees with the same session name — presence looked stale from each side, and sends queued where nobody was reading. A `--wait-for drained` receipt timing out was the only visible symptom. (This is the incident behind discussions #202/#203.)

This PR fixes the root cause and adds two small delivery-visibility improvements that came out of the same incident:

- **`coop exec` pins the session root to an absolute path** before starting wake and exporting `AM_ROOT`/`AM_BASE_ROOT`. The session's mailbox identity is frozen at launch instead of drifting with cwd.
- **`amq env` shell output (plain and `--export`) emits absolute roots** — these exports exist to pin a terminal, and a relative export re-splits per cwd. JSON output is unchanged.
- **`amq reply --wait-for <stage> --wait-timeout <d>`** — parity with `amq send --wait-for`, so reply delivery can be confirmed instead of assumed. In the incident, the reply leg was exactly where the split re-appeared.
- **`amq who` text output prints a `Base root:` header** naming the tree it reads, so per-root active/stale presence isn't mistaken for global liveness. JSON output unchanged (kept as a bare array for compat).
- **Test-isolation fix:** `TestResolveEnvConfigAutoDetect` and `TestResolveEnvConfigNoConfig` read the developer's real `~/.amqrc`/`AMQ_GLOBAL_ROOT` and failed on any machine that has one; they now isolate `HOME` and `AMQ_GLOBAL_ROOT`.

## Compatibility

- Sessions launched from a project root behave identically — the same directory is used, just addressed absolutely.
- If anyone relied on a relative `AM_ROOT` deliberately re-resolving per cwd, this changes that; I couldn't find a case where that's desirable rather than a footgun, but happy to gate it behind a flag if you know of one.
- `env --json`, `who --json` unchanged.

## Tests

- New `internal/cli/roots_absolute_test.go`: absolute pinning (relative→abs under a temp cwd, absolute unchanged), env shell/export absolute emission, reply `--wait-for drained` matched-receipt flow (mirrors the send cross-session wait test harness), invalid-stage rejection, who base-root header.
- `go test ./...`: 748 passed, 0 failed — including on a machine with a real `~/.amqrc`, which previously broke the two env tests above.
- `gofmt`/`go vet` clean; I don't have golangci-lint locally, so relying on CI for lint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+
+- `amq reply --wait-for <stage> --wait-timeout <duration>` blocks on the
+  recipient's delivery receipt, mirroring `amq send --wait-for`, so reply
+  delivery can be confirmed instead of assumed.
+- `amq who` text output now prints a `Base root:` header naming the tree it is
+  reading, so per-root `active`/`stale` presence is not mistaken for global
+  liveness when the same session name exists in another root (JSON output is
+  unchanged).
+
+### Fixed
+
+- `amq coop exec` now pins the session root to an absolute path before
+  starting wake and exporting `AM_ROOT`/`AM_BASE_ROOT`. A relative root
+  (e.g. from the `.agent-mail` default) re-resolved against every future cwd
+  of the agent process, silently splitting one session name into
+  per-directory mailbox trees across git worktrees — peers on the "same"
+  session could not see each other and sends queued where nobody was reading.
+- `amq env` shell output (plain and `--export`) now emits absolute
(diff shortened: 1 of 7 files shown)


avivsinai (approved):
Reviewed the session-root, env, reply receipt, and who-output changes. Local verification: targeted internal/cli tests for absolute roots/reply wait/who output passed, go test ./... passed, and go vet ./... passed. GitHub CI, Gitleaks, and GitGuardian are green on head 423cbc3021edb1e06aaa0fea377c3d448db76c3a.

PR #232, opened 2026-07-12
fix: harden Darwin wake lifecycle and root classification

ohade:
## Summary
Harden AMQ wake lifecycle handling on Darwin and fix default `.agent-mail` base-root classification when unrelated sibling `agents/` directories exist.

## Changes
- stabilize exact Darwin wake-target reuse and treat zombie wake processes as stopped
- add identity-safe `amq wake retire` for an exact managed wake, including PID/start-token/boot-ID and lock revalidation before signaling
- preserve mailbox wake-target metadata while safely retiring the managed process
- classify the default `.agent-mail` base before the sibling-session heuristic, preventing `$HOME/.claude/agents` from making same-tree sends look cross-tree
- add regression coverage for target mismatch, identity changes, missing locks, stale locks, zombie processes, and poisoned sibling directories

## Testing
- [ ] `make ci` passes locally (only blocked because `golangci-lint` is not installed on this machine)
- [x] `make check-skills fmt-check vet test smoke`
- [x] new focused Darwin wake-retire and root-classification tests

## Related Issues
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ editing the release PR.
 - Darwin wake identity now uses the stable `kern.bootsessionuuid` instead of
   the wall-clock-derived `kern.boottime`, while retaining a bounded legacy
   compatibility path and a safe `kern.boottime` fallback for older macOS.
+  Darwin process inspection also treats unreaped `SZOMB` wake processes as
+  stopped, so their stale wake locks can be replaced.
 - Accept-existing wake startup now verifies that an existing wake's injection
   mode and, for external wakes, persisted injector path and ordered arguments
   match the requested target; mismatched, owner-orphaned, or concurrently
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,10 @@ editing the release PR.
   mode and, for external wakes, persisted injector path and ordered arguments
   match the requested target; mismatched, owner-orphaned, or concurrently
   replaced terminal targets do not report readiness.
+- Add `amq wake retire` for identity-verified managed shutdown of an exact
+  inject-via wake while preserving its mailbox and saved target. The command
+  refuses target mismatches, unverified locks, and process/lock replacement
+  races before signaling.
 
 ## [0.41.1] - 2026-07-10
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,16 @@ editing the release PR.
   documented Dependabot-detection pattern, which stays `dependabot[bot]`
   regardless of who updates the branch.
 
+### Fixed
+
+- Darwin wake identity now uses the stable `kern.bootsessionuuid` instead of
+  the wall-clock-derived `kern.boottime`, while retaining a bounded legacy
+  compatibility path and a safe `kern.boottime` fallback for older macOS.
+- Accept-existing wake startup now verifies that an existing wake's injection
+  mode and, for external wakes, persisted injector path and ordered arguments
+  match the requested target; mismatched, owner-orphaned, or concurrently
+  replaced terminal targets do not report readiness.
+
 ## [0.41.1] - 2026-07-10
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ amq doctor --ops
 amq doctor --ops --json
 amq doctor --ops --fix-wake-locks
 amq wake repair --me codex
+amq wake retire --me codex --inject-via /absolute/injector \
+  --inject-arg inject --inject-arg cmux --inject-arg cmux:surface:<uuid>
 ```
 
 Wake locks reported by `doctor --ops` can be `stale`, `unverified`, or, in JSON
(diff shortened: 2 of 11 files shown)


avivsinai:
Thanks for this — the Darwin wake work is genuinely good, and the `kern.bootsessionuuid` and `SZOMB` findings are real. We want it. Four changes before it can land, and one of them is a decision we'd already reached independently.

## 1. Drop the `classifyRoot` commit and rebase on `main` — #231 has landed

#231 landed the root-classification fix on `main` in [774f568](https://github.com/avivsinai/agent-message-queue/commit/774f568efc656a3c57ed7c4c48ee5022f4e415d9), so your `fix: classify default AMQ base before sibling scan` now conflicts with `main` on `internal/cli/common.go`. Please rebase and drop that commit — it's entirely orthogonal to your wake work.

Worth knowing *why* we didn't take your version, because it turned out to matter: your guard keys on the literal directory name,

```go
if filepath.Base(filepath.Clean(root)) == defaultCoopRoot { return "" }
```

and that short-circuit had a **cross-tree guard bypass**. A session directory literally named `.agent-mail`, nested under a `.agent-mail` base (`.agent-mail/.agent-mail/`), gets promoted from session to base because the basename check runs before the "parent is `.agent-mail`" evidence — `sameBaseTree` flips, and the unqualified explicit-root send guard stops refusing a send that escapes (via symlink) into a physically different queue. An adversarial review pass (GPT-5 Pro) caught it; we reproduced it, plus a root-local-`.amqrc` and a case-insensitive-spelling variant. The landed fix runs the parent-convention check before the basename check (in both `classifyRoot` and `baseRootOf`) and recognizes the default root by filesystem identity (`os.SameFile`) so case-insensitive spellings can't evade it. It also fixes the custom-`.amqrc`-base case your version missed. Deeper pre-existing hardening is tracked in #233. So: just rebase and drop the commit; `main` has the complete fix.

## 2. Drop the `CHANGELOG.md` edit entirely

Two problems. Release Please owns `CHANGELOG.md` — it generates entries from conventional squash commits on `main`, and there is no changelog gate in `ci.yml`, so the manual edit isn't needed. More importantly, the block landed **inside the already-released `0.42.0` section**, immediately above `## [0.41.1]`, so it rewrites the notes of a shipped release.

If you need richer release notes than the PR title gives, use `BEGIN_COMMIT_OVERRIDE` in the PR body instead.

## 3. Retitle to `feat(wake):`

The repo is squash-only, so the PR title becomes the conventional commit on `main`. This branch contains `feat(wake): retire exact managed wake safely` — a new CLI command — under a `fix:` title, which would ship a new feature as a patch release. Something like `feat(wake): harden managed wake lifecycle` fits.

## 4. Keep `wake retire` in this PR — but document its real boundary

We pushed back hard on whether `retire` should exist at all (this project is deliberately minimal, and wake *supervision* is explicitly a launchd/systemd concern). It survives, on evidence: `requireWakeLockUsable` on `main` already tells operators to *"stop the existing wake and retry"* while providing no identity-safe way to do it. `repair` only replaces a proven-**stale** inject-via wake; `doctor --ops --fix-wake-locks` only removes proven-**stale** locks; a `wakeLockValid` + identity-confirmed wake with a mismatched mode/target is neither. So the gap pre-exists, and your (c) — the accept-existing target/args check — *widens* it. That makes `retire` the remediation for what (c) newly refuses, so the two belong in the same change. Don't split it out.

Two asks:

- **Docs**: add `retire` to the README wake section and to the command list in `CLAUDE.md`.
- **Don't overclaim the boundary.** `retire` requires an exact saved inject-via target, so it does **not** stop raw/paste/`none` wakes and does not resolve every `--inject-mode none` mismatch. Please state it precisely:
  - `repair` — proven-stale inject-via target → remove stale lock, restart
  - `doctor --ops --fix-wake-locks` — proven-stale lock → remove only
  - `retire` — exact inject-via target → stop an identity-confirmed live wake, no restart, mailbox and saved target preserved
  - launchd/systemd/shell still owns raw-wake stop, and decides *when* any managed wake stops

---

Net: this becomes `feat(wake):` containing the Darwin identity fix, the zombie fix, the accept-existing target check, and `retire` — with the changelog and `classifyRoot` changes removed. Happy to re-review as soon as it's pushed.

avivsinai:
Landed the sound, self-contained portion of this PR — the Darwin `bootsessionuuid` identity + `SZOMB` zombie handling — as #236 (merged to main in ae776b8), crediting you as co-author.

Two parts were deliberately **not** landed and are tracked separately:
- `amq wake retire` + the accept-existing target reuse: an adversarial security review (GPT-5 Pro) showed the underlying wake lock/signal machinery isn't yet a safe foundation for an "identity-safe" stop command (bare-PID signal that can hit a recycled PID, non-atomic lock lifecycle, mode persistence). That hardening — and reintroducing `retire` on top — is #235.
- The root-classification commit: superseded by #231 (already on main), which fixed the same area including a cross-tree guard bypass found during review.
- The manual CHANGELOG edit: Release Please owns the changelog.

Thanks for the contribution — the Darwin identity/zombie work was solid and is shipping. Closing in favor of #236.

Issue #235, opened 2026-07-13
Wake-lifecycle signal safety: stable process capability, tri-state inspection, atomic lock lifecycle (prereq for wake retire)


avivsinai:
## Summary

The wake lock/signal machinery (`internal/cli/wake_unix.go`, `wake_lock.go`) identifies and stops wake processes by **PID plus a non-atomic identity recheck**, and mutates the `.wake.lock` with **non-atomic compare-then-act** sequences. That is adequate for today's best-effort orphan cleanup, but it is not a foundation on which to build an operator-facing, "identity-safe" stop command. An `amq wake retire` command was designed and reviewed, but deferred out of PR #232 precisely because these weaknesses make its advertised guarantees ("identity-safe", "PID reuse fails closed") untrue.

These weaknesses are **pre-existing on `main`** — `terminateWakeProcess`, `signalWakeProcess`, `wakeProcessStillMatches`, `removeWakeLockIfUnchanged`, and `requireWakeLockUsable` all exist today, and `terminateAndRemoveOrphanedWakeLock` already drives the signal path during owner-orphan wake replacement. Surfaced by an adversarial review (GPT-5 Pro) and reproduced at the source level.

## Weaknesses

1. **PID-reuse race in the signal path.** `terminateWakeProcess` sends `SIGTERM`, waits ~100ms, then `SIGKILL`, all by bare PID after a non-atomic identity check. If the verified wake exits and the kernel recycles its PID to another same-UID process between the last check and the signal, the unrelated process is signaled — and `SIGKILL` is uncatchable. `os.FindProcess` on Linux may obtain a pidfd, but only *after* validation, so it can bind the replacement.

2. **Mode compatibility is not exact** (`requireWakeLockUsable`). Only the `requiredMode == none && lock != none` direction is checked; an existing `none` wake therefore satisfies a requested raw/paste/auto. Worse, raw/paste/auto modes are not persisted in the lock at all, so exact compatibility is impossible with the current schema.

3. **Post-`SIGTERM` inspection failure is treated as proof of termination.** `wakeProcessStillMatches` returns a boolean; `false` conflates "proven gone" with "inspection failed / identity temporarily unavailable". On Darwin, `kill(pid,0)` can succeed while `SysctlKinfoProc` fails, yielding `false` even though the wake may still be running — so a stop "succeeds" and the lock is removed while the process lives, enabling duplicate injection.

[...]

PR #241, opened 2026-07-19
Fix exact wake-mode compatibility

ohade:
## Summary

Persist the normalized wake injection mode in every wake lock and require exact mode compatibility when `--accept-existing-wake` reuses a live wake.

## Why

An existing `none` wake could satisfy a requested raw/paste wake, and raw/paste/auto modes were not persisted, so exact compatibility could not be enforced. This is the narrow prerequisite identified in #235.

## Scope

- Persist `none`, `raw`, `paste`, the effective auto mode, and `inject-via` in the lock.
- Reject `none` for non-`none` requests.
- Reject raw/paste mismatches.
- Add regression coverage for both compatibility cases.

This PR deliberately does not reintroduce `amq wake retire`; PID-reuse-safe signaling, tri-state inspection, and atomic lifecycle work remain outside this change.

## Validation

- `go test ./...`
- `go test ./internal/cli`

### DIFF
--- a/internal/cli/wake_unix.go
+++ b/internal/cli/wake_unix.go
@@ -124,8 +124,8 @@ createLock:
 	if options.target != nil {
 		lock.WakeMode = wakeTargetInjectVia
 		lock.TargetDigest = wakeTargetDigest(*options.target)
-	} else if options.wakeMode == wakeInjectModeNone {
-		lock.WakeMode = wakeInjectModeNone
+	} else {
+		lock.WakeMode = options.wakeMode
 	}
 	if hostname, err := os.Hostname(); err == nil {
 		lock.Hostname = hostname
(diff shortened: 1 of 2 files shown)


avivsinai (changes requested):
The core fix is right and we want it — `auto` is resolved to its effective raw/paste before both persisting and comparing (`wake_unix.go:791-796`), so the lock never lies about actual behavior, and the `TargetDigest != ""` requirement in `wakeLockHasUsableNotificationPath` is correct hardening. Requesting changes on three items (independent Claude + Codex passes, converged):

## 1. Legacy-lock upgrade regression (blocking)

Old binaries persisted `WakeMode` only when it was `none` — the exact line this PR removes (`wake_unix.go:124-129`) — so every pre-upgrade raw/paste/auto lock carries `WakeMode:""`. The new exact check (`wake_unix.go:246`) rejects `"" != "raw"/"paste"` even when the mode matches, so `coop exec --require-wake` (which routes through `--accept-existing-wake`) fails against a healthy pre-upgrade wake until that wake is restarted. With no safe `wake retire`, a forced restart is a real upgrade regression — the too-strict direction #235 explicitly warned about. Darwin detail that shapes the fix: `inspectWakeProcessPlatform` doesn't populate `Args`, so the legacy wake's true mode can't be reconstructed from the process.

Agreed matrix: **empty persisted `WakeMode` = legacy TTY non-none → accept a normalized raw or paste request, reject `none` and `inject-via`; non-empty → exact equality.** The raw-vs-paste ambiguity then exists only for already-running old binaries and ages out as they restart.

## 2. Tests don't pin the actual behavior change (blocking)

Both added tests hand-set `Lock.WakeMode` and call `requireWakeLockUsable` directly. Nothing tests the persistence wiring — revert `lockWakeMode = effectiveInjectMode(...)` to persisting literal `"auto"` and the suite stays green while the original bug returns. Please add:
- Run-level round-trip: `--inject-mode auto` + `--me claude` → lock persists `WakeMode:"raw"`; a non-claude/codex handle → `"paste"`.
- The four legacy cases against a `WakeMode:""` lock: raw accepted, paste accepted, `none` rejected, `inject-via` rejected.

## 3. Retitle (CI)

The only CI failure is `check_pr_title.py`; the Go job never ran. Retitle to `fix(wake): enforce exact wake-mode compatibility` and CI will rerun. (Full suite passes locally at 9a10cf1.)

Nit: `wake_lock.go:28` comment still says `WakeMode` is "inject-via or none".

Follow-up (not this PR — will get its own issue under #235): the reuse path never compares inject-via injector identity/`TargetDigest`, so an existing inject-via wake with a different injector satisfies a request. Pre-existing on main; your `omitempty` schema handling is fine both directions.

avivsinai:
Superseded by #246, now merged as 8153b8a — your commit was cherry-picked with authorship preserved and you're co-authored on the squash commit. #246 adds the review items on top: legacy pre-persistence lock fallback (empty `wake_mode` = legacy TTY wake → accepts matching raw/paste, rejects `none`/`inject-via`), run-level persistence regression pinning the `effectiveInjectMode` wiring, and the four legacy matrix tests. Thanks for the fix and the clean #235-scoped framing — it's exactly the prerequisite it claimed to be.

PR #265, opened 2026-07-21
fix(wake): ignore stale backlog on managed startup

ohade:
## Summary
- add a wake startup baseline that suppresses only messages already present in inbox/new
- enable it automatically for coop exec and wake repair
- preserve stale messages as unread without emitting delivery receipts
- keep messages arriving after the snapshot eligible for notification

## Why
A managed wake could inject an old unread notification, including an urgent Ctrl-C, while a newly launched agent terminal was still starting. Baseline-before-readiness prevents that stale backlog from becoming terminal input.

## Verification
- make ci
- focused baseline/readiness and unread/no-receipt regression tests
- gitleaks staged scan

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,7 @@ amq dlq list --me <agent> [--new | --cur] [--json]
 amq dlq read --me <agent> --id <dlq_id> [--session <name>] [--ignore-session-pin] [--json]
 amq dlq retry --me <agent> --id <dlq_id> [--session <name>] [--ignore-session-pin] [--all] [--force]
 amq dlq purge --me <agent> [--session <name>] [--ignore-session-pin] [--older-than <duration>] [--dry-run] [--yes]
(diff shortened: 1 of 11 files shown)


avivsinai:
Landed as #267 (merged to main) — thank you, Ohad. We took the work over with your encouragement and you're credited as co-author on the commit.

Summary of what changed from your original: we kept the core idea (`--baseline-existing` snapshot-before-inject) and hardened it substantially through review — switched suppression from filename to dev+inode+ctime instance identity (so DLQ redelivery stays notify-eligible), fixed the fresh-mailbox and 32-bit Linux build paths, and **descoped** the cross-process reuse handoff after a ChatGPT-Pro review showed it couldn't guarantee raw-TIOCSTI terminal-queue safety without destructive input flushing. The reuse case is now an explicitly-documented residual. It went through 4 ChatGPT-Pro gate rounds converging from 4 P1 races to clean, with no message data loss at any point.

Really appreciate you surfacing the stale-injection bug — it was real and worth fixing properly. Closing this in favor of #267.

PR #266, opened 2026-07-22
fix(wake): defer managed wake until SessionStart

ohade:
## Summary
Defers managed wake startup until agent SessionStart while preserving arrivals between the pre-launch floor capture and notifier readiness. This avoids binding a wake to a stale or restored terminal identity and fails visibly without consuming queued messages.

## Changes
- add `coop exec --defer-wake` with a private, identity-bound unread-floor manifest
- add `wake --baseline-file` with exact floor persistence, reuse, and generation-bound readiness
- start the watcher before catch-up, rotate floors only for the same owner and transport, and preserve exact repair/retire behavior
- canonicalize macOS control-socket roots so `/var` and `/private/var` address the same wake
- document the companion SessionStart hook contract

## Testing
- [x] `make ci` passes
- [x] New baseline, rotation, downtime, lifecycle, and Darwin root-alias tests added
- [x] Paired installed-binary E2E with `amq-keepalive` verified exact `notifier_live`, no receipts, and exact retirement
- [x] `gitleaks` found no leaks

## Related Issues
Supersedes #265.

Companion integration: ohade/amq-keepalive#1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,18 @@ the wake stores a repair target so `amq wake repair` can restart wake later
 without restarting the agent TUI. If `--require-wake` reuses an existing wake,
 that wake must already have repair metadata to be repairable.
 
+SessionStart-managed launchers can use `amq coop exec --defer-wake ...` to start
+the TUI before attaching wake. AMQ captures the pre-launch unread floor in a
+private manifest and exports its path as `AMQ_WAKE_BASELINE_FILE`; the hook must
+start `amq wake --baseline-file "$AMQ_WAKE_BASELINE_FILE"` and require its
+generation-bound readiness acknowledgement. Messages arriving between capture
+and hook attachment remain eligible for notification. Baseline capture failure
+does not block the TUI: AMQ exports the sanitized `AMQ_WAKE_BASELINE_ERROR`, and
+the hook should show a visible “wake unavailable; messages remain queued”
+warning. Installing that SessionStart hook is a hard prerequisite: without it,
+`--defer-wake` intentionally starts the TUI but cannot attach wake or show the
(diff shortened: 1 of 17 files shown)


avivsinai:
Closing as superseded, per the discussion in the #235 thread — reasoning on the record.

The launch-gap problem is real and the watcher-before-catch-up sequencing is sound. But this adds a second on-disk baseline mechanism (`--baseline-file` + manifest) alongside the `--baseline-existing` mechanism merged in #267, and the two use different suppression identities: this one keys on message-ID first, #267 keys on device+inode+ctime. That divergence is a concrete regression — a message baselined here, moved to DLQ, and retried is silently suppressed for the baseline generation (it stays queued and drainable, but never notifies), because `fsq.RetryFromDLQ` preserves the header ID on a new queue occurrence. #267's inode+ctime identity treats the retry as a new instance and notifies; reproduced both ways.

Built-in `coop exec` already prepares the fenced #267 wake before exec, so in-tree the launch gap is closed; the remaining case is the external-supervisor shape. We don't want two baseline protocols with two identities in core's most fragile subsystem.

You did surface a genuine core bug though: `wake repair` relaunches with `--baseline-existing` and re-snapshots `inbox/new`, so a message arriving during notifier downtime is reclassified as repair-start backlog and its notification suppressed for the repaired generation (still drainable, not lost). That's ours to fix in core, on the same inode+ctime identity as #267 — tracked as a follow-up, starting from a failing mainline regression.

The ownership/reclaim work in the stacked #270 is the part we want in core; we're taking that in-tree ourselves as a fresh PR co-authored with you. Thanks for the contribution — it moved #235 forward.

PR #270, opened 2026-07-23
Safely reclaim abandoned wake ownership

ohade:
## Summary

- capture the unread-message baseline before launching the agent and allow a SessionStart hook to attach wake afterward
- bind wake ownership and transitions to an exact process identity, reclaiming a handle only when its prior owner is conclusively dead
- serialize competing same-handle claims while leaving independent handles and mailbox trees isolated
- hide conclusively dead owner-bound sessions from default discovery while retaining `amq who --all`

## Why

Reusing an abandoned AMQ session could leave stale wake ownership behind. A newly launched Claude or Codex process could then fail to attach wake, or compete with another launcher for the same handle.

This change makes takeover lazy and conservative: it occurs during connection, under the existing per-handle lifecycle lock, and fails closed whenever ownership is live, legacy, incomplete, corrupt, or otherwise ambiguous.

## Validation

- `make ci`
- `GOWORK=off go test -race ./...`
- `gitleaks dir . --no-banner --redact --exit-code 1`
- outgoing-history secret and private-path scan
- five-reviewer public-contribution audit

## Known follow-ups

This is intentionally a draft. The public audit identified work to complete before review:

- preserve retryability when interrupt-key injection fails but the notice succeeds
- define bounded retry behavior for transient startup and steady-state injection failures
- avoid scanning the receipt directory once per unread message
- fail closed when baseline cleanup cannot read the current wake target
- raise changed-line coverage from approximately 75.5% to at least 80%


avivsinai (changes requested):
Not merge/release ready. I confirmed the draft's listed follow-ups are real blockers on head `bd6ab0941037f03b6813fd9e1f3264fa1dff17e2`.

1. `internal/cli/wake.go:235-249`: an interrupt-key injection failure is ignored. If the following notice injection succeeds, the urgent messages are marked seen, so their interrupt semantics are lost permanently. Preserve retryability when the key fails.
2. `internal/cli/wake_unix.go:1479-1483,1539-1548`: startup injection failure exits immediately, while a steady-state failure clears `pendingNotify` and only logs. With no later filesystem event, the unread message is never retried. Add bounded backoff/retry with an explicit terminal failure policy.
3. `internal/cli/wake_baseline.go:257-265`: any `readWakeTarget` error is treated like “not referenced,” after which a valid baseline may be deleted. Cleanup must fail closed on unreadable/corrupt/ambiguous target state.
4. `internal/cli/wake.go:193-194,304-317`: every unread message calls `receipt.List`, rescanning the receipt directory once per message. Load the drained receipt IDs once per notification scan.
5. `internal/cli/wake_owner_claim_unix_test.go:724-736`: the goroutines call a helper that can invoke `t.Fatal`, but the parent does two unbounded receives. A setup/validation failure causes a suite-wide timeout instead of an actionable failure. Return errors through the result channel and bound the receives.

Repository gates also remain open: the PR is intentionally draft, the title must be conventional (for example `fix(wake): reclaim abandoned wake ownership`), GitHub CI stopped at the title check and skipped every build/test step, and the PR body says changed-line coverage is still below the intended 80% floor.

Local evidence from a secure detached worktree: `make ci` passes, `go test -race ./...` passes, and `git diff --check` passes. Those are necessary but do not resolve the correctness findings above.

internal/cli/coop_exec_test.go
```diff
@@ -133,6 +135,143 @@ func TestCoopExecRequireWakeRejectsNoWake(t *testing.T) {
...
+	})
+	t.Setenv(envWakeOwner, `{"pid":999,"process_start":"inherited","boot_id":"old-boot"}`)
+
+	oldExec := coopExecProcess
+	defer func() { coopExecProcess = oldExec }()
+	sentinel := errors.New("exec intercepted")
+	coopExecProcess = func(_ string, _ []string, env []string) error {
+		path := envValue(env, envWakeBaselineFile)
+		if path == "" || envValue(env, envWakeBaselineError) != "" {
+			t.Fatalf("deferred env missing manifest or contains error: %#v", env)
```
avivsinai, line 188: Blocking: this failure path formats the entire `env` slice inherited from `os.Environ()`. A failed test therefore publishes every ambient credential to CI logs (the diagnostic run reproduced this with real secrets). Assert/redact only the specific AMQ keys needed by the test; never print the full environment.

avivsinai (commented):
Additional blocking security finding from the failure-path test run.

avivsinai:
Superseded by #277, which is now merged as `84ea950`.

The owner-lifecycle work from this PR is what we wanted in core, and it has landed — rebuilt fresh from `main` (ported behavior and tests rather than rebasing, so none of #266's baseline fields came along), and credited to your design and implementation via `Co-Authored-By`.

What landed in #277: exact owner identity (PID + process start + boot id + session binding), a per-handle lifecycle guard spanning inspection and mutation, tri-state owner classification with automatic takeover only of a conclusively-dead owner, fail-closed handling of ambiguous identity, the exact owner handoff through `coop exec`, an atomic `0400` mixed-version-fenced claim that older `amq` binaries preserve rather than mutate, and a guarded `amq wake recover-owner` so a stranded claim always has a deliberate recovery path.

It went through a frozen design cleared by ChatGPT-Pro, a six-way parallel adversarial code review, and a ChatGPT-Pro code gate that caught a PID-reuse signaling defect — now fixed and covered by regressions asserting no signal ever reaches a reused PID.

`who --all` dead-session visibility and the repair-time notification-continuity fix remain tracked as separate follow-ups. Thanks for the contribution — it moved #235 forward.

Issue #278, opened 2026-07-24
wake repair loses notification continuity: downtime arrivals reclassified as startup backlog


avivsinai:
Follow-up from the #277 owner-lifecycle work; confirmed in code review (ChatGPT-Pro owner-lens + Claude + Codex, 2026-07-23) and deliberately excluded from that PR's scope.

## Bug

`wake repair` relaunches the notifier with `--baseline-existing` (`buildRepairWakeArgs`, internal/cli/wake_unix.go), which re-snapshots `inbox/new` at repair time. A message that arrived **while the notifier was dead** is present in `inbox/new` at that moment, gets captured into the fresh baseline, and is reclassified as startup backlog — its notification is suppressed for the repaired generation, and repeated repairs can perpetuate the suppression indefinitely.

**Not data loss**: the message stays queued and drainable. Only the notification is suppressed — but for an urgent/interrupt message, that defeats the point of wake.

## Direction (agreed with the Pro-gated design review)

One baseline model, two capture times — do NOT reintroduce a second suppression identity:
1. Start with a **failing mainline regression**: start an owned wake and establish its baseline → kill the notifier while the owner stays alive → deliver an urgent message during downtime → `wake repair` → assert the downtime message notifies. (Expected to fail today.)
2. Then an **internally persisted floor** using the SAME dev+inode+ctime instance identity as #267, bound to the exact owner + wake generation, reused only for repair of that same live owner, never after takeover or reboot-driven owner invalidation. Internal implementation detail — not a file/env contract handed to external tools.

Refs: #267 (baseline-existing), #277 (owner lifecycle), #266 (closed; its repair design was evidence this concern is real).

avivsinai:
Resolved by PR #287 (squash commit `327f47469ef7008c7d3c0d8d133d2a3778d9ae86`) and shipped in v0.47.1 (release commit `418931297294f6cea9ac3fc0f17e9325b3977339`). The released tree includes `TestWakeRepairNotifiesMessageDeliveredDuringDowntime` plus exact-identity, lineage-bound, authority-bound repair-floor coverage, so a message arriving during notifier downtime is no longer re-baselined away by repair. Closing as released.

PR #290, opened 2026-07-25
fix(wake): bind repair inbox parent authority

avivsinai:
## Summary
- open `agent/inbox/new` component-by-component without following an intermediate symlink
- bind the original inbox-parent identity into the private repair handoff schema and digest
- reject parent replacement during admission and the repaired wake runtime

## Verification
- focused lifecycle tests (`count=50`)
- focused race tests (`count=10`)
- `go test ./...` and `go test -race ./...`
- `go vet ./...`
- Linux test compile/vet and Windows/FreeBSD builds

Issue: #278

Merge is held for exact-head ChatGPT Pro review.


avivsinai:
Closing unmerged. GO 3 changed generic wake admission so an unverified ownerless lock no longer blocks recovery; the original wake-repair case no longer justifies this repair-specific hardening surface. The repair path will be evaluated for retirement in a separate design packet.

PR #292, opened 2026-07-25
fix(wake): validate private bootstrap fds atomically

avivsinai:
## Summary
- capture and scrub every private wake bootstrap environment value before parsing
- validate the complete Darwin six-FD transaction, including pairwise alias rejection, before adoption
- prepare all descriptor semantics before creating file owners, watchers, goroutines, or the wake loop
- roll back the whole validated inherited tuple on preparation failure
- scrub and reject unsupported private descriptor state on Linux

## Verification
- focused tests x10 and focused race tests
- full suite and full race suite
- vet, formatting, native and Linux lint
- Darwin arm64 and Linux amd64 test cross-builds

Merge is held for exact-head ChatGPT Pro review.

### DIFF
--- a/internal/cli/wake_owner_child_darwin.go
+++ b/internal/cli/wake_owner_child_darwin.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"strconv"
-	"strings"
 	"sync"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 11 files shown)


avivsinai:
Closing unmerged. GO 3 changed generic wake admission so an unverified ownerless lock no longer blocks recovery; the original wake-repair case no longer justifies this repair-specific hardening surface. The repair path will be evaluated for retirement in a separate design packet.

PR #420, opened 2026-08-03
fix(wake): guard readiness publication on the capability path

### LINKED ISSUE #419: wake: bare `amq wake` crashes at startup — empty ready-file path reaches os.Rename (regression in v0.49.11)
## Symptom

A bare `amq wake` — no `--ready-file` — dies within ~300ms of start:

```
install wake ready file: rename .wake-metadata.tmp.<pid>.<nanos>.0 : no such file or directory
```

Note the rename destination is the empty string. The wake never acquires its
notifier role, so agents launched behind it go silently unreachable: messages
still land in the maildir, but nothing is ever injected.

This is unconditional on the bare-wake path. It reproduces with `--inject-via`
and with `--inject-mode none`, with and without a controlling TTY, on a fresh
`amq init` root and on a long-lived `coop init` root, with `--root` and with
`AM_ROOT`.

## Version

Reproduced from source at `bfbbdd4` (v0.52.0), macOS 27.0 (arm64),
go1.26.5 darwin/arm64.

Bisected against released darwin/arm64 binaries — fresh root per arm, 3s
liveness check:

| version | bare `amq wake` |
|---|---|
| 0.48.0 | alive |
| 0.49.9 | alive |
| 0.49.10 | alive |
| **0.49.11** | **dead** |
| 0.50.1 | dead |
| 0.51.1 | dead |
| 0.52.0 | dead |

## Reproduce

```bash
go build -o /tmp/amq ./cmd/amq
root=$(mktemp -d)
AMQ_NO_UPDATE_CHECK=1 /tmp/amq init --root "$root" --agents alice,bob
AMQ_NO_UPDATE_CHECK=1 /tmp/amq wake --me alice --root "$root" --inject-mode none
```

Expected: wake starts and blocks, watching the inbox.
Actual: exits immediately with the message above.

## Root cause

`onPrepared` in `internal/cli/wake_unix.go` (v0.52.0 line 2242) publishes
readiness through the agent-dir-capability variant:

```go
return writeWakeReadyFileAgainstOwnerInDir(
        activeAgentDir, root, me, readyFile, currentWake, requestedOwner)
```

The empty-path guard lives only in the outer wrapper
`writeWakeReadyFileAgainstOwner` (`internal/cli/wake_ready_unix.go:33`):

```go
if path == "" {
        return nil
}
```

`…InDir` has no such guard, so an empty `readyFile` reaches
`writeWakeGenerationFile("")` → `writeWakeMetadataFile("")` →
`os.Rename(tmp, "")`, which is ENOENT. The tmp basename in the error is the
`"wake-metadata"` fallback from `createWakeMetadataTempFile`, which is only
selected when `filepath.Base(path)` has no non-dot characters — i.e. it is
itself evidence that `path` was empty.

v0.49.10 called the guarded wrapper at the same site:

```go
return writeWakeReadyFileAgainstOwner(
        root, me, readyFile, currentWake, requestedOwner)
```

The switch to the `…InDir` variant landed in `27923be9`
("fix(wake): preserve retry delivery guarantees", released as v0.49.11) and
left the guard behind in the wrapper.

`writeWakeReadyFileAgainstOwner` and `writeWakeReadyFile` now have no
non-test callers, which is why the orphaned guard is invisible: every
production call goes through `…InDir`.

## Why CI and coop consumers don't see it

`--ready-file` is an internal flag supplied on exactly two paths — the
repair-wake child (`buildRepairWakeArgs`) and `coop exec`
(`coop_exec_unix.go:699`). Anything launched through `amq coop exec` therefore
always has a non-empty ready path and never reaches the bug.

The consumers that break are the ones spawning `amq wake` directly, which the
README documents for exactly this case:

```bash
amq wake --me orchestrator \
  --inject-via /absolute/path/to/ghostty-bridge \
  --inject-arg exec \
  --inject-arg "$TERMINAL_ID"
```

That is how I hit it: an orchestrator spawns one detached `amq wake` per worker
and one for the driver. After upgrading to 0.52.0 every wake in every session
died at startup while the spawning side reported success, so worker
notification and reply injection both stopped with no error surfaced anywhere
except the wake log.

## Note

Happy to send a PR — I have the fix and a regression test at the
`runWakeWithLoop` seam alongside the existing
`TestRunWakeWithLoopWritesReadyFileAfterLock`.

est7:
Fixes #419.

## What was broken

A bare `amq wake` — no `--ready-file` — died within ~300ms of start:

```
install wake ready file: rename .wake-metadata.tmp.<pid>.<nanos>.0 : no such file or directory
```

The rename destination is the empty string. Regression bisected to v0.49.11;
v0.49.10 and earlier are fine.

## Why

`onPrepared` publishes readiness through the agent-dir-capability variant:

```go
return writeWakeReadyFileAgainstOwnerInDir(
        activeAgentDir, root, me, readyFile, currentWake, requestedOwner)
```

but the `if path == "" { return nil }` guard lived only in the outer
`writeWakeReadyFileAgainstOwner`. `27923be9` moved this call site onto the
capability variant to reuse the already-open agent directory, and the guard
stayed behind in the wrapper. An empty `readyFile` then reached
`os.Rename(tmp, "")`.

`writeWakeReadyFileAgainstOwner` and `writeWakeReadyFile` have no remaining
non-test callers, so the guard was sitting on a path production no longer
takes.

`--ready-file` is supplied only by the repair-wake child and by `coop exec`, so
anything launched through `amq coop exec` never reached it. Consumers that
spawn `amq wake` directly — the `--inject-via` orchestrator invocation in the
README — lost notification delivery entirely, with the failure visible only in
the wake log.

## The change

The guard moves to `writeWakeReadyFileAgainstOwnerInDir`, which is the one
point every caller passes through, so entering with an already-open capability
cannot bypass it. Nothing else about publication, validation, or ownership
changes.

## Test

`TestRunWakeWithLoopPreparesWithoutReadyFile`, at the same `runWakeWithLoop`
seam as the existing `TestRunWakeWithLoopWritesReadyFileAfterLock`: bare wake
must complete `onPrepared` and still publish its generation-bound prepared
marker.

Red before the fix, with the production error verbatim:

```
--- FAIL: TestRunWakeWithLoopPreparesWithoutReadyFile (0.04s)
    wake_unix_test.go:1015: prepare wake without --ready-file: install wake ready file: rename .wake-metadata.tmp.65885.1785730150077823000.0 : no such file or directory
```

Also verified end to end on the built binary: the repro command from #419 exits
immediately before the change and stays alive with clean stderr after it.

## Gates

macOS 27.0 arm64, go1.26.5. `check-skills`, `fmt-check`, `vet`, `lint`
(0 issues) and `smoke` all pass.

`make test` has three failures on my machine, all in
`internal/cli/worktree_divergence_test.go` and all reproducing identically on
an unmodified checkout of this commit's parent: their fixtures run
`git add .amqrc`, and my global `core.excludesfile` ignores `.amqrc`, so the
add is refused. Environment-specific and unrelated to this change — flagging it
rather than touching those tests.

### DIFF
--- a/internal/cli/wake_ready_unix.go
+++ b/internal/cli/wake_ready_unix.go
@@ -30,9 +30,6 @@ func writeWakeReadyFileAgainstOwner(
 	expected wakeLockInspection,
 	requestedOwner *wakeOwner,
 ) error {
-	if path == "" {
-		return nil
-	}
 	agentDir, err := openWakeAgentDir(root, me)
 	if err != nil {
 		return err
@@ -56,6 +53,14 @@ func writeWakeReadyFileAgainstOwnerInDir(
 	expected wakeLockInspection,
 	requestedOwner *wakeOwner,
 ) error {
+	// No readiness path means no supervising parent is waiting on the
+	// handshake — a bare `amq wake` rather than a repair child or a
+	// `coop exec` launch. Publishing is then a no-op, and this is the one
+	// place every caller passes through, so the rule cannot be bypassed by
+	// entering with an already-open agent directory capability.
+	if path == "" {
+		return nil
+	}
 	if agentDir == nil {
 		return fmt.Errorf("wake agent directory capability is missing")
 	}
--- a/internal/cli/wake_unix_test.go
+++ b/internal/cli/wake_unix_test.go
@@ -991,6 +991,42 @@ func TestRunWakeWithLoopWritesReadyFileAfterLock(t *testing.T) {
 	}
 }
 
+// A bare `amq wake` — the invocation an orchestrator spawns directly, with no
+// supervising parent waiting on a readiness handshake — carries no
+// --ready-file. Preparation must still complete: readiness publication is the
+// handshake's side of the contract, not a precondition for owning the wake.
+func TestRunWakeWithLoopPreparesWithoutReadyFile(t *testing.T) {
+	root := secureTempDirForTest(t)
+	if err := fsq.EnsureRootDirs(root); err != nil {
+		t.Fatalf("EnsureRootDirs: %v", err)
+	}
+	if err := fsq.EnsureAgentDirs(root, "orchestrator"); err != nil {
+		t.Fatalf("EnsureAgentDirs: %v", err)
+	}
+
+	injector := writeExecutableForTest(t, "injector")
+	errDone := errors.New("done")
+	err := runWakeWithLoop([]string{
+		"--root", root,
+		"--me", "orchestrator",
+		"--inject-via", injector,
+	}, func(cfg wakeConfig) error {
+		if err := cfg.onPrepared(nil); err != nil {
+			t.Fatalf("prepare wake without --ready-file: %v", err)
+		}
+		if _, exists, preparedErr := readWakeGenerationFile(
+			wakePreparedPath(root, "orchestrator"),
+			"wake prepared marker",
+		); preparedErr != nil || !exists {
+			t.Fatalf("generation-bound prepared marker missing: exists=%v err=%v", exists, preparedErr)
+		}
+		return errDone
+	})
+	if !errors.Is(err, errDone) {
+		t.Fatalf("expected loop sentinel error, got %v", err)
+	}
+}
+
 func TestRunWakeWithLoopBaselinesBeforeReadiness(t *testing.T) {
 	root := secureTempDirForTest(t)
 	if err := fsq.EnsureRootDirs(root); err != nil {


avivsinai:
Merged — thank you @est7. Verified independently before landing: your test reproduces the exact #419 rename-ENOENT death at base and passes with the guard moved to the single choke point every caller uses; also live-smoked a bare `amq wake` end to end (prepared marker, injector fallback, clean SIGTERM). Ships in the next 0.52.x patch.

PR #422, opened 2026-08-03
fix(wake): retry the retained-inbox kqueue wait on EINTR

### LINKED ISSUE #421: wake: EINTR kills the retained-inbox kqueue watcher, so one message injects several doorbells
## Symptom

One delivered message produces several wake injections. In the run that surfaced
this, a single reply — one message ID, confirmed sent once by the sending agent —
injected the notification **three times**. The third injection arrived *after* the
recipient had already drained the mailbox, so it woke the agent against an empty
inbox.

The wake diagnostic log shows exactly two extra events for that one message:

```
amq wake: wake inbox watcher failed: wait for retained wake directory event: interrupted system call; retrying watcher setup
amq wake: wake inbox watcher failed: wait for retained wake directory event: interrupted system call; retrying watcher setup
```

One delivery + two watcher restarts = three injections.

## Root cause

`interrupted system call` is `EINTR`. The retained-inbox kqueue loop treats it as
fatal:

https://github.com/avivsinai/agent-message-queue/blob/main/internal/cli/wake_repair_dirs_darwin.go#L114-L125

```go
count, err := unix.Kevent(w.kqueueFD, nil, events, nil)
if err != nil {
	select {
	case <-w.closing:
		return
	default:
		w.fail(fmt.Errorf("wait for retained wake directory event: %w", err))
		return
	}
}
```

Any error kills the watcher goroutine. But the Go runtime delivers `SIGURG` to
threads for asynchronous preemption, and `golang.org/x/sys/unix.Kevent` does not
retry on `EINTR`, so a long-lived blocking `Kevent` gets interrupted as a matter
of course. The watcher dies for a reason that is not an error.

This loop is the outlier. Every other blocking syscall in the package retries:
`wake_owner_observation_darwin.go:175,206,275`, `wake_owner_observation_linux.go:148`,
`wake_tiocsti_unix.go:100,209`, `wake_terminate_linux.go:282`. The closest one,
`monitorDarwinWakeOwnerObservation`, is the same `unix.Kevent` loop shape and does
handle it:

```go
count, err := unix.Kevent(capability.kqueueFD, nil, events, nil)
if errors.Is(err, syscall.EINTR) {
	continue
}
```

The duplicate injection then comes from the restart path, which is doing the right
thing given a genuine watcher failure — `retryWatcher` in `wake_unix.go` clears
`cfg.baselineExisting` and sets `pendingNotify = true`, i.e. "I may have missed
events, so re-scan and notify". That fail-open choice is correct on its own and I
am not proposing to change it; it just converts every spurious EINTR into a
spurious doorbell.

Darwin only. The Linux watcher is a separate implementation
(`wake_repair_dirs_linux.go`) and is not affected.

## Why this is showing up now

It is not new — the loop dates to 327f474 (2026-07-24) and the restart path to
27923be9 (2026-07-30). It was unreachable in practice because bare `amq wake`
died at startup (#419), so no wake lived long enough to be interrupted. With that
fixed, wake stays up, and this is the next failure along the same path.

## Impact

Agents driven by wake injections get woken repeatedly for a single message. In an
orchestrated setup each spurious doorbell costs a full agent turn, and a wake
against an already-drained mailbox is indistinguishable to the agent from a real
notification.

## Version

Reproduced on `main` at bfbbdd4 (v0.52.0), macOS 27.0 (arm64), go1.26.5
darwin/arm64. The observed run used a build of #419's branch, which is what allows
bare wake to stay up long enough to hit this.

## Fix

Retry on `EINTR` in the retained-inbox kqueue loop, matching the six existing
call sites. PR follows, with a regression test that injects `EINTR` through a
package-level wait seam and asserts the watcher survives and still forwards the
inbox event.

est7:
## Summary

The retained-inbox kqueue watcher treated every `unix.Kevent` error as fatal, so
an `EINTR` killed the watcher goroutine. Since the Go runtime delivers `SIGURG`
for asynchronous preemption and `x/sys/unix.Kevent` does not retry, a long-lived
blocking wait gets interrupted as a matter of course — the watcher dies for a
reason that is not an error.

`retryWatcher` then does the right thing for a genuine watcher failure (clear the
baseline, set `pendingNotify`, re-scan and notify), which turns each spurious
death into a spurious doorbell. In the run that surfaced this, one delivered
message injected three wake notifications, the last one after the recipient had
already drained the mailbox.

I deliberately did **not** touch `retryWatcher`. Re-notifying after a watcher
restart is the correct fail-open choice on its own; it was only pathological
because EINTR made restarts constant.

Darwin only — the Linux watcher is a separate implementation.

## Changes

- `wake_repair_dirs_darwin.go`: retry the retained-inbox wait on `EINTR`, matching
  the six existing call sites in this package. The closest one,
  `monitorDarwinWakeOwnerObservation`, is the same `unix.Kevent` loop shape and
  already handles it — this loop was the outlier.
- The wait goes through a package-level `waitRetainedWakeInboxEvent` seam so the
  interrupt can be injected deterministically, following the
  `writeDarwinWakeOwnerObservationCancel` precedent in
  `wake_owner_observation_darwin_test.go`.
- Preserved the existing precedence: a wait error during `Close()` still returns
  through the `w.closing` branch before the retry is considered.

## Testing

- [x] New test added: `TestDarwinRetainedWakeWatcherRetriesInterruptedWait`
  injects `EINTR` on the first wait, then asserts the watcher survives, still
  forwards the normalized inbox event, and performed at least two waits.
- [x] Reverse-probed. Reverting to the exact pre-fix state (retry removed, both
  imports dropped) fails the new test with `retained watcher closed before
  forwarding message trigger` — the same channel-close the production bug causes.
  It passes again after restoration.
- [x] `go test ./internal/cli/ -run TestDarwinRetainedWakeWatcher -count=1` — all
  5 tests pass.
- [x] `go vet ./internal/cli/`, `gofmt -l internal/cli/`, `go build ./...` — all
  clean.
- [ ] `make ci` — not clean in my environment, but not because of this change:
  `TestDoctorOpsHintsPerWorktreeSessionForLocalRootSources`,
  `TestDoctorOpsWarnsWhenPeerPresenceIsFresherInSameSessionOtherWorktree`, and
  `TestDoctorOpsDoesNotWarnForOlderPeerOrFresherCallerPresence` fail identically
  on a clean `origin/main` tree here (verified by stashing this change and
  re-running), and they appear to be sensitive to local git configuration —
  the run emits git's `advice.addIgnoredFile` hint. Flagging rather than
  claiming a green I did not get.

## Related Issues

Fixes #421

### DIFF
--- a/internal/cli/wake_repair_dirs_darwin.go
+++ b/internal/cli/wake_repair_dirs_darwin.go
@@ -3,13 +3,21 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 	"sync"
+	"syscall"
 
 	"github.com/fsnotify/fsnotify"
 	"golang.org/x/sys/unix"
 )
+
+// waitRetainedWakeInboxEvent is a seam so tests can inject the interrupts the
+// Go runtime delivers in production.
+var waitRetainedWakeInboxEvent = func(kqueueFD int, events []unix.Kevent_t) (int, error) {
+	return unix.Kevent(kqueueFD, nil, events, nil)
+}
 
 type retainedWakeInboxKqueueWatcher struct {
 	kqueueFD  int
@@ -113,15 +121,18 @@ func (w *retainedWakeInboxKqueueWatcher) run() {
 	eventName := filepath.Join(w.authority.inboxPath, "retained-inbox-event.md")
 	for {
 		events := make([]unix.Kevent_t, 2)
-		count, err := unix.Kevent(w.kqueueFD, nil, events, nil)
+		count, err := waitRetainedWakeInboxEvent(w.kqueueFD, events)
 		if err != nil {
 			select {
 			case <-w.closing:
 				return
 			default:
-				w.fail(fmt.Errorf("wait for retained wake directory event: %w", err))
-				return
 			}
+			if errors.Is(err, syscall.EINTR) {
+				continue
+			}
+			w.fail(fmt.Errorf("wait for retained wake directory event: %w", err))
+			return
 		}
 		if count == 0 {
 			continue
--- a/internal/cli/wake_repair_dirs_darwin_test.go
+++ b/internal/cli/wake_repair_dirs_darwin_test.go
@@ -6,12 +6,15 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
+	"syscall"
 	"testing"
 	"time"
 
 	"github.com/avivsinai/agent-message-queue/internal/format"
 	"github.com/avivsinai/agent-message-queue/internal/fsq"
 	"github.com/fsnotify/fsnotify"
+	"golang.org/x/sys/unix"
 )
 
 type darwinRetainedWakeWatcherFixture struct {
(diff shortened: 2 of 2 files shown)


avivsinai:
Merged — thank you @est7. Verified independently before landing: your test fails behaviorally at base with the exact EINTR symptom and passes with the fix; all five production `unix.Kevent` sites were re-inspected (the other waits already retried EINTR; this was the last one). Ships in the next 0.52.x patch.

PR #424, opened 2026-08-03
feat(wake): acknowledge injected doorbells with --retry-until injected

### LINKED ISSUE #423: wake: exec injectors cannot acknowledge doorbell presentation; unchanged cohorts re-prompt busy agents
## Problem

`amq wake --inject-via <presenter>` exposes two different delivery facts that today collapse into one:

1. **Presentation**: the injector exited zero, proving the terminal accepted the complete doorbell.
2. **Consumption**: a message left `inbox/new`, proving durable consumer progress (the `drained` receipt).

The retry ladder only recognizes fact 2. When the consumer is an interactive agent (Codex/Claude in a Ghostty pane) that is booting or in the middle of a long thinking turn, the inbox stays unchanged past the retry deadline and the same doorbell is injected again — the agent sees one message as two, three, or four prompts, each burning a full agent turn.

This stacked with #419 and #421 in a live incident (each independently multiplied prompts), but it is a distinct gap: for standalone use "retry until drained" is the right, safe default; for an integration that already owns receipt-based recovery there is simply no way to tell wake "the presenter accepted it; stop reannouncing this cohort".

## Proposal

Add an explicit acknowledgement boundary, opt-in and additive:

```
--retry-until drained|injected   # default: drained (behavior unchanged)
```

Under `injected` (requires `--inject-via`):

- a zero-exit, non-attention injector completion acknowledges exactly the unchanged physical inbox cohort and removes its retry deadline;
- injector failure and timeout stay retryable;
- attention/fallback outcomes are never acknowledgements;
- a new or in-place-replaced physical message re-arms one fresh doorbell;
- acknowledgement never creates or replaces a `drained` receipt — consumption is still proven only by the original message's receipt.

The policy joins injector target identity and persists as an additive `retry_until` field, omitted for the default so existing wake targets and state documents keep byte-identical serialization; wake repair and retire propagate it.

PR incoming with this shape.

## Alternative considered: acknowledgement as data instead of a phase

A structurally cleaner model would drop the extra doorbell phase entirely: keep the machine at Idle/Retrying, and track an identity-keyed (`device+inode+ctime`) `acked` cohort as plain data. `plan()` would compute `pending = current − acked`, announce only when pending is non-empty, and set `acked = current` on successful injection. The same-name in-place-replacement edge disappears structurally — a replaced file has a new identity, so it lands in `pending` with no special-casing.

The PR deliberately does **not** do this: the existing doorbell machine is phase-based and also drives the recovery and attention ladders, so the data-model rewrite would touch every phase consumer for an edge that maildir's unique filenames make nearly unreachable. The minimal version adds one phase, handles replacement with an explicit predicate plus a regression test, and leaves the default path untouched. If a second acknowledgement source ever appears, the acked-set model is the natural refactor to revisit.

## Compatibility note

`retry_until` is additive: an older binary reading newer state silently ignores the field and falls back to `drained` semantics. Integrations that rely on `injected` should pin the amq binary they spawn (e.g. project an absolute path into the presenter environment) rather than trusting PATH.

est7:
## Summary

Adds an explicit presentation-acknowledgement boundary to the wake doorbell: `--retry-until drained|injected` (default `drained`, behavior unchanged). Under `injected`, a zero-exit, non-attention `--inject-via` completion acknowledges the unchanged physical inbox cohort and stops reannouncing it, so an interactive agent that is booting or mid-turn no longer receives the same doorbell as multiple prompts.

Mail consumption is still proven only by the original message's `drained` receipt; injection acknowledgement never creates or replaces a receipt. Design discussion and the rejected acked-set alternative: #423.

## Current composition and exact review identity

Composed onto released `v0.53.0` main; the prior stacked-branch note no longer applies.

- Base: `d05110343763fa9cd7a6668611a96e8816912cde`
- Head: `981f03a58cc3b68854fbde109a4340a58113e6b8`
- Tree: `7e93c792ddb821918182fcbafc03a0923df271ee`
- Plain diff SHA-256: `d3af92dda86c9516d6741dc19b17837b3a75a710e9ed71509930209fb1d82751`
- Full-index binary diff SHA-256: `c665721f1d58e538d0c6e4057fbc326d50c09941844ba74b44e003779f99ec0d`

All local validation below binds to that exact tree.

## Changes

- Adds `--retry-until drained|injected` to `amq wake`; `injected` requires `--inject-via`.
- Adds the `wakeDoorbellAnnounced` phase: an unchanged cohort stays suppressed after successful non-attention injection; injector failure, timeout, and attention outcomes retain the existing retry ladders.
- A new message or an in-place replacement with a changed physical identity re-arms one fresh doorbell.
- Persists `retry_until` in injector identity. The default remains omitted for byte-compatible existing target/state documents; omitted and explicit `drained` compare equivalently.
- Validates stored values canonically and rejects malformed case/whitespace variants.
- Propagates the policy through repair and retire; retire refuses a mismatched acknowledgement policy.
- Documents the contract in `docs/wake-doorbell-acknowledgement.md`, README, and the wake lifecycle docs.

## Validation

Platform: `Darwin 25.5.0 arm64`; `go1.26.5 darwin/arm64`.

- [x] `make ci`
- [x] Focused race suite covering canonical stored-policy validation, target byte/digest goldens, state round-trip, retire mismatch refusal, injector identity, and CLI normalization
- [x] Exact non-default target JSON golden and digest; default serialization/digest remains unchanged
- [x] Negative validation for unknown, mixed-case, and whitespace-padded stored policies
- [x] Current-head hosted GitHub checks: Linux CI run 31097229875, Gitleaks, and GitGuardian passed

Historical live evidence: the original head `0472d7238bb52a2da3d2d6b87b726110853fb03c` was exercised against a Ghostty-pane orchestrator and produced one prompt for one message while busy, without a late-drain duplicate. That live exercise was not rerun at the composed head and is not presented as current-head or production-gate evidence. AMQ/cmux production gates are not claimed here.

## Release semantics

This is a backward-compatible user-visible feature. The conventional `feat` title should make Release Please propose `v0.54.0` from `v0.53.0`; no manual tag or patch-version override is intended.

## Related issues

Closes #423

TestSameWakeInjectorIdentityUsesOnlyPathAndOrderedArgs: renamed to TestSameWakeInjectorIdentityUsesPathOrderedArgsAndRetryPolicy; the original path and ordered-args identity assertions are preserved, and the test now also covers the additive retry_until field in injector identity.


avivsinai (dismissed):
Reviewed at head 83fb2f41ba393e4f4f47c691263309d4bf37ca57 (tree d35d6cf7b859afd8867ba932a90ef123dda90156). Local make ci and focused race validation pass; stored-policy canonicalization, exact serialization/digest goldens, persistence, and retire identity mismatch behavior reviewed. Historical live proof remains explicitly non-current; AMQ/cmux production gates are not claimed.

avivsinai (approved):
Reviewed at head 981f03a58cc3b68854fbde109a4340a58113e6b8 (tree 7e93c792ddb821918182fcbafc03a0923df271ee). The only follow-up changes the failing Linux test fixture from t.TempDir to the existing secureTempDirForTest helper; production validation is unchanged. Focused package tests and independent race test pass. Awaiting hosted Linux confirmation.

PR #478, opened 2026-08-09
fix(wake): reuse detached external injectors safely

ohade:
## Problem

AMQ v0.60.0 rejects a healthy keepalive-managed `--inject-via` wake on re-entry when the detached process has no controlling TTY. The wake can already have a valid lock-bound target, but the generic terminal-orphan check rejects it before those stronger facts are validated.

The companion keepalive path also hid pre-readiness wake stderr behind `exit status 1`, lacked reliable plain-build version provenance, and was being developed in more than one repository.

## Changes

- Reuse a detached external-injector wake only after validating the exact persisted target, target digest, root, agent, executable, and ordered arguments.
- Preserve replacement for terminal wakes and same-TTY/different-session wakes.
- Keep keepalive-managed wakes intentionally ownerless instead of inheriting ambient `AMQ_WAKE_OWNER`.
- Preserve up to 16 KiB of pre-readiness stderr through a bounded detached drain helper and keep an established wake alive if that optional reader disappears.
- Recheck cancellation immediately before starting a durable wake.
- Add `-v`, `--version`, and `version`, including release stamping and Go VCS metadata fallback.
- Make this repository the public source of truth for keepalive code and documentation.

The launcher initialization ordering and `reattach --no-start` changes were implemented and tested separately. This PR does not change cmux surface identity handling.

## Verification

At exact head `bc169ee69a8fde80c36b09f621683dd83721a25e`:

- GitHub CI: build, cross-platform compile, tests, smoke test, lint, and both secret scans passed.
- `env -u AMQ_KEEPALIVE_DISABLED make ci` passed locally.
- `go test -race ./internal/keepalive/amq ./internal/cli -count=1` passed.
- The Linux retained-state replacement regression passed in a fresh Linux container.
- Companion co-op launcher, bootstrap, and project-launcher suites passed against an AMQ binary built from the reviewed candidate.
- Official v0.60.0 archive provenance verification passed.
- Three independent audit reviewers completed; all validated findings were resolved. Two additional Claude lanes produced no review because their quota was exhausted and were not used as evidence.

Status is **SOURCE PASS / LIVE PENDING**. Production-cmux Gates 1–5 were not run because this PR did not install binaries or mutate the user's production cmux state.

## Requested reproduction

Please repeat Gal's original macOS re-entry scenario against this exact head and report:

- AMQ, keepalive, and cmux versions;
- first attach and same-target re-entry results;
- `amq wake check --json` before re-entry;
- whether the target remained unique and lock-bound;
- whether any raw-wake fallback was needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,30 @@ For installations made with the install script or another manual binary install:
 amq upgrade
 ```
 
+### Keepalive companion
+
+`amq-keepalive` is developed and released from this repository alongside AMQ.
+`make build` produces both binaries, and each AMQ release includes a separate
+`amq-keepalive` archive stamped with the same release version. Verify a build
+with any equivalent form:
+
+```bash
+amq-keepalive -v
+amq-keepalive --version
+amq-keepalive version
+```
+
+Keepalive-managed wakes deliberately discard an ambient `AMQ_WAKE_OWNER` token
+so they remain ownerless and can outlive the short launcher that registered
+them. During startup, keepalive captures at most 16 KiB of child stderr in a
+private regular file and includes it in a pre-readiness exit error. This keeps
+actionable AMQ diagnostics visible without attaching a long-lived wake to the
+caller's terminal or to a pipe whose reader exits first.
+
+See [docs/amq-keepalive.md](docs/amq-keepalive.md) for adapters, launcher
+ordering, supervision, retirement, owner recovery, and the detached stderr
+drain contract.
+
 ## Quick Start
 
 ### 1. Initialize Project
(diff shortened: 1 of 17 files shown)


avivsinai (approved):
Two independent reviews (Claude + Codex via AMQ, thread review/pr-478) at head bc169ee found no blocking issues. Reuse gate verified fail-closed: accept-existing path validates persisted target, digest-to-lock binding, root/agent, and exact injector path + ordered argv; mismatches refuse readiness. Live macOS exercise: exact-target reuse returned readiness, different injector failed without a ready file, exact retire stopped only the original PID. go test -race passed across internal/keepalive and internal/cli wake matrices (19+ iterations; one unreproducible local flake did not reproduce in 19 retries on PR head and 4 runs on main).

PR #649, opened 2026-08-26
fix(coop): detect named inject TTY via /dev/tty

bnsd55:
## Summary
- `--named` TUI inject (`codex` / Cursor `agent`) always warned `no controlling terminal` because the helper starts with stdin on `/dev/null` and the ready check used `tiocsti.IsTTY()` (stdin).
- Probe `/dev/tty` instead, which is what TIOCSTI injection already opens. Stdin stays disconnected so the helper does not steal the agent's TTY.

## Test plan
- [ ] `go test ./internal/cli -run 'CoopNamed|InjectCoopNamed'`
- [ ] `amq coop exec --named --me cursor agent` no longer prints `no controlling terminal`
- [ ] Claude/Pi argv `--name` path unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/cli/coop_exec_named_unix.go
+++ b/internal/cli/coop_exec_named_unix.go
@@ -16,8 +16,11 @@ import (
 const coopNamedTUIStartupDelay = 3 * time.Second
 
 var (
+	openCoopNamedTTY = func() (*os.File, error) {
+		return os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	}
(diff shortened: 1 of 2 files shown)


avivsinai:
Thanks @bnsd55 — correct diagnosis and the right owning-layer fix (the helper's stdin is deliberately /dev/null, so probing /dev/tty is exactly what the TIOCSTI path already relies on). Merged as a32c500; ships in v0.70.1.

PR #676, opened 2026-08-27
fix(cli): isolate nested worktree root discovery

### LINKED ISSUE #648: amq-squad adoption asks on v0.70.0: base-root discovery override, scoped-pattern allowedTools, approvals_reviewer seeding (follow-up to #557)
@avivsinai — following your close-out invitation on #557 ("Reopen here or open a fresh issue with your findings and we take them in the same gate"). This covers the three remaining items after our **v0.70.0** re-verification: 4 of 7 findings verified fixed, and the seeded Codex flags confirmed accepted with our exact value pairs. Continuation of #557; original design in #480.

Ordered by severity. None of these blocks our v2.30.0 adoption milestone — it ships additive and opt-in — but item 1 must close before we point a real launch at the substrate.

1. **Base-root discovery overrides caller intent (state-corruption class).** Add `--base-root`
   to the CLI `--plan` path, and/or make an explicit `.amq/launch.json` in cwd suppress upward
   base discovery. Highest severity: it writes to the wrong live queue rather than refusing.
   *Evidence:* t13 finding (f) + incident record on `task/t13` — reproduced by tripping it, not
   inferred. Nested scratch project under a parent repo; `setup` `written` list omitted `amqrc`
   (up-tree config adopted); `Apply` created a stray session in the parent repo's live base root
   (11 files, 0 messages, live session intact, verified by both fullstack and lead).
2. **Claude scoped-pattern `--allowedTools` support (privilege-escalation-or-regression fork).**
   Accept `Bash(<prefix>:*)`. Today the only accepted substitute for a narrow grant is blanket
   `Bash`, so a consumer must choose between escalating privilege and dropping the grant.
   *Evidence:* v0.70.0 value-grammar probes recorded on `task/t15` — ACCEPT `Bash`, `Bash,Read`,
   `Edit`, `mcp__x__y`; REJECT `Bash(ls)`, `Bash(ls:*)`, `Bash:*`, `Bash(gh pr create:*)`.
   Separately worth reporting as a strictness inversion: the same grammar **accepts**
   `["--allowedTools","--dangerously-skip-permissions"]` and `["--allowedTools","--verbose"]` —
   permissive about flag-looking values, strict about real Claude pattern syntax.
3. **`-c approvals_reviewer` seeding (refuse-to-launch class).** Add the key with an
   `auto_review` allowlist, and specify raw value bytes — our real emission carries literal TOML
   quotes (`agent_defaults.go:17`).
   *Evidence:* t13 section (b) — both `approvals_reviewer="auto_review"` (quoted, our live
   default) and `approvals_reviewer=auto_review` (bare) reject, so the key is unlisted rather
   than the quoting being wrong; the full live approve-for-me set rejects as a whole and accepts
   with only this key removed. Matches the t12 freeze answer.

**Non-blocking, worth carrying:** CLI/package parity — the CLI `--plan` path has no
`--base-root` or `--caller-context` flag and rejects a full `PrepareRequestV1`
(`unknown field "request_version"`), so findings H.1 and `caller_context` are reachable only
from the Go package. Acceptable for us since we adopt via the package, but it contradicts the
frozen §11 claim that CLI and package share one contract.

**And what the design got right** — recorded deliberately, not as filler. Across three
verification rounds (v0.63.1, v0.65.2, v0.70.0) everything actually *in* the contract behaved
exactly as the frozen text said, on first release each time. Specifically worth naming:
- The §4 trust boundary refuses a provider executable located inside the repository
  (`provider_project_contained`) — caught our own fake-provider test harness, correctly.
- `executable_identity` binding is genuinely hard to get right and is right: a same-canonical-path
  inode swap moves `subject_digest` while leaving `plan_digest` and `trust_digest` stable, and
  `Apply` then refuses with `subject_changed` and zero mutation. The counterexample from our own
  earlier review no longer reproduces.
- Exact-root emission holds across sibling worktrees — one absolute `--root`, no `--session`,
  identical across seats with differing cwd. This was our hard adoption gate and it has held on
  every release tested.
- Contract stability: 0.61.1 with all 12 features intact from v0.65.2 through v0.70.0,
  additive-only, no feature removed and no version bumped, through five minors that shipped
  substantial unrelated work (ACP companion, bridge courier, Grok peer, coop `--named`).

---
Sent by Claude on behalf of Omri. Evaluation by my amq-squad team: senior-dev (Codex) and fullstack (Claude Code).

avivsinai:
#663 closed #648 for `amq setup --project-root` (explicit beats inferred; parent `.amqrc` ignored). Omri on v0.73.0 still reported the same discovery seam for a **nested git worktree** / nested checkout that is itself a worktree, not a scratch subdirectory: cwd under a parent that already has `.amqrc` could adopt the parent live base root.

## What was actually left

`.amqrc` / `.agent-mail` walk-up already stopped at the innermost `.git`. The leftover was the **relative-root** walk (`resolveRoot` → `findRootInParents`) and `findAmqrcForRoot`, which had no git ceiling. A relative `.agent-mail` or a queue path inside a nested/linked worktree could still resolve to the parent live queue.

Scratch subdirs inside the **same** worktree are unchanged (v0.61 / #663 implicit case). A nested or linked worktree is its own ceiling: own config wins, otherwise fail closed. No new config format.

## Changes

- Share `gitWorktreeRootFrom` and apply that ceiling to `findRootInParents` and `findAmqrcForRoot`
- **5381d96 (review):** `EvalSymlinks` the walk start when a ceiling exists, matching `findAndLoadAmqrc` / `detectAgentMailDir`, so a logical `$PWD` through a symlink cannot skip the break and adopt the parent live queue
- Regression: nested independent checkout, nested linked worktree, and a **symlinked path component** into the nested worktree do not adopt the parent live base root
- Existing #663 `--project-root` / symlink-refuse tests kept

## Cloud VM named check (5381d96)

```text
$ go test ./internal/cli -count=1 -timeout 30s -v -run 'TestNestedWorktreeDoesNotAdoptParentLiveBaseRoot'
--- PASS: TestNestedWorktreeDoesNotAdoptParentLiveBaseRoot (0.32s)
    --- PASS: .../symlinked_path_into_nested_worktree_does_not_adopt_parent (0.04s)
PASS
ok  	github.com/avivsinai/agent-message-queue/internal/cli	0.320s
```

Does not stack on #673 / #674 / #675. Does not merge or tag a release.

<div><a href="https://cursor.com/agents/bc-d267c47a-be74-4a14-9c97-3dddac7d21b4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d267c47a-be74-4a14-9c97-3dddac7d21b4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/internal/cli/common.go
+++ b/internal/cli/common.go
@@ -417,10 +417,17 @@ func absPath(path string) string {
 
 func findRootInParents(startDir, relative string) (string, bool) {
 	dir := startDir
+	ceiling := ""
+	if top, insideGit := gitWorktreeRootFrom(startDir); insideGit {
+		ceiling = top
+	}
 	for {
 		candidate := filepath.Join(dir, relative)
 		if dirExists(candidate) {
 			return absPath(candidate), true
+		}
+		if ceiling != "" && sameCleanPath(dir, ceiling) {
+			break
 		}
 		parent := filepath.Dir(dir)
 		if parent == dir {
--- a/internal/cli/common.go
+++ b/internal/cli/common.go
@@ -416,11 +416,14 @@ func absPath(path string) string {
 }
 
 func findRootInParents(startDir, relative string) (string, bool) {
-	dir := startDir
 	ceiling := ""
 	if top, insideGit := gitWorktreeRootFrom(startDir); insideGit {
 		ceiling = top
+		if resolved, resolveErr := filepath.EvalSymlinks(startDir); resolveErr == nil {
+			startDir = resolved
+		}
 	}
+	dir := startDir
 	for {
 		candidate := filepath.Join(dir, relative)
 		if dirExists(candidate) {
--- a/internal/cli/env.go
+++ b/internal/cli/env.go
@@ -590,11 +590,27 @@ func gitWorktreeRootFromCWD() (string, bool) {
 	if err != nil {
 		return "", false
 	}
-	if resolved, resolveErr := filepath.EvalSymlinks(cwd); resolveErr == nil {
-		cwd = resolved
+	return gitWorktreeRootFrom(cwd)
+}
+
+// gitWorktreeRootFrom returns the innermost Git worktree or bare-repository
+// boundary that contains path. A nested or linked worktree is its own
+// ceiling, so a parent checkout's .amqrc or .agent-mail is not inferred.
+func gitWorktreeRootFrom(path string) (string, bool) {
+	path = strings.TrimSpace(path)
+	if path == "" {
+		return "", false
+	}
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return "", false
+	}
+	path = abs
+	if resolved, resolveErr := filepath.EvalSymlinks(path); resolveErr == nil {
+		path = resolved
 	}
 	bareCandidate := ""
-	for dir := cwd; ; dir = filepath.Dir(dir) {
+	for dir := path; ; dir = filepath.Dir(dir) {
 		marker := filepath.Join(dir, ".git")
 		_, statErr := gitMarkerLstat(marker)
 		if statErr == nil {
@@ -1075,8 +1091,12 @@ func findAmqrcForRoot(root string) (amqrcResult, error) {
 		if absErr != nil {
 			return amqrcResult{}, absErr
 		}
+		ceiling := ""
+		if top, insideGit := gitWorktreeRootFrom(absRoot); insideGit {
+			ceiling = top
+		}
 		dir := absRoot
-		for !isHomeConfigDir(dir) {
+		for ceiling != "" || !isHomeConfigDir(dir) {
 			rcPath := filepath.Join(dir, ".amqrc")
 			if info, statErr := amqrcLstat(rcPath); statErr == nil {
 				if err := validateAmqrcInfo(rcPath, info); err != nil {
@@ -1100,6 +1120,9 @@ func findAmqrcForRoot(root string) (amqrcResult, error) {
 			if !os.IsNotExist(readErr) {
 				// Permission or I/O error — report it, don't mask it.
 				return amqrcResult{}, fmt.Errorf("cannot read .amqrc at %s: %w", rcPath, readErr)
+			}
+			if ceiling != "" && sameCleanPath(dir, ceiling) {
+				break
 			}
 			parent := filepath.Dir(dir)
 			if parent == dir {
--- a/internal/cli/env.go
+++ b/internal/cli/env.go
@@ -1094,6 +1094,9 @@ func findAmqrcForRoot(root string) (amqrcResult, error) {
 		ceiling := ""
 		if top, insideGit := gitWorktreeRootFrom(absRoot); insideGit {
 			ceiling = top
+			if resolved, resolveErr := filepath.EvalSymlinks(absRoot); resolveErr == nil {
+				absRoot = resolved
+			}
 		}
 		dir := absRoot
 		for ceiling != "" || !isHomeConfigDir(dir) {
--- a/COOP.md
+++ b/COOP.md
@@ -105,7 +105,9 @@ Equivalent explicit root form: `--root .agent-mail/<session>`.
 
 That isolation also applies across git worktrees. A relative project root such
 as `{"root":".agent-mail"}` resolves separately inside every worktree, even when
-the session names match. To share a mailbox intentionally, configure the same
+the session names match. A nested or linked worktree under a parent that
+already has `.amqrc` uses its own config or fails closed; it does not adopt
+the parent live queue. To share a mailbox intentionally, configure the same
 absolute `.amqrc` root in each worktree, or remove the relative project config
 and set `AMQ_GLOBAL_ROOT` to one absolute base. Use `amq doctor --ops` when a
 delivery receipt times out; it can name divergent same-session roots when a
--- a/docs/launch-api.md
+++ b/docs/launch-api.md
@@ -118,7 +118,9 @@ directory that already owns its config is the project root. A parent `.amqrc`
 found above an explicit project root (or a cwd that owns its config) is
 ignored, not adopted, so the parent tree stays byte-identical; the implicit
 case still refuses an upward-discovered parent `.amqrc` with the parent path
-named.
+named. A nested or linked Git worktree is the same ceiling without the flag:
+discovery uses that worktree's own config or fails closed, and does not adopt
+the parent live base root.
 
 ```json
 {
--- a/docs/session-routing.md
+++ b/docs/session-routing.md
@@ -10,9 +10,11 @@ exact root and `AM_SESSION` is empty. Do not set the identity tokens manually.
 
 Root precedence is explicit `--root`, `AM_ROOT`, project `.amqrc`,
 `AMQ_GLOBAL_ROOT`, then eligible implicit fallbacks. Within a Git checkout,
-only repo-local implicit state is eligible. An unreadable or invalid project
-`.amqrc` blocks lower-precedence fallback; an explicit `--root` or `AM_ROOT`
-can override it intentionally.
+only repo-local implicit state is eligible. A nested or linked Git worktree
+is its own discovery ceiling: a parent checkout's `.amqrc` or `.agent-mail`
+is not adopted. An unreadable or invalid project `.amqrc` blocks
+lower-precedence fallback; an explicit `--root` or `AM_ROOT` can override it
+intentionally.
 
 A direct `--root` selects a queue. It is not a federation route. `send` refuses
 an explicit root in a different base tree when the caller has an active
--- a/internal/cli/nested_worktree_discovery_test.go
+++ b/internal/cli/nested_worktree_discovery_test.go
@@ -96,6 +96,52 @@ func TestNestedWorktreeDoesNotAdoptParentLiveBaseRoot(t *testing.T) {
 		assertDoesNotAdoptParentLiveRoot(t, parent, parentMail, nested)
 	})
 
+	t.Run("symlinked path into nested worktree does not adopt parent", func(t *testing.T) {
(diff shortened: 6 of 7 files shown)


internal/cli/common.go
```diff
@@ -417,11 +417,18 @@ func absPath(path string) string {
...
+		ceiling = top
+	}
 	for {
 		candidate := filepath.Join(dir, relative)
 		if dirExists(candidate) {
 			return absPath(candidate), true
 		}
+		if ceiling != "" && sameCleanPath(dir, ceiling) {
+			break
+		}
```
avivsinai, line 434: **Blocking: the ceiling is defeated by a symlinked path component — parent live-queue adoption survives.**

`gitWorktreeRootFrom` resolves symlinks before computing the ceiling, but the walk here uses the *unresolved* `startDir`, and `sameCleanPath` is `Abs`+`Clean` only (no symlink resolution). When `startDir` contains a symlinked component, `sameCleanPath(dir, ceiling)` never matches, the break never fires, and the walk degrades to the old unbounded behavior — silently adopting the parent `.agent-mail`.

This is reachable in production: `resolveRoot` passes `os.Getwd()`, which prefers logical `$PWD` when it names the cwd, and any shell `cd` through a symlink (aliased checkout dirs, macOS `/tmp` → `/private/tmp`) leaves the symlinked path in `$PWD`.

Verified with a probe on this branch (parent repo with `.amqrc`+`.agent-mail`, nested independent checkout, `parent/alias → parent/nested`, cwd entered via the alias so `$PWD` is logical):

```text
resolveRoot(".agent-mail") from symlinked cwd adopted parent live queue .../parent/.agent-mail
findRootInParents via symlinked path escaped the ceiling: .../parent/.agent-mail
```

The repo already has the alignment pattern for exactly this: `findAndLoadAmqrc` and `detectAgentMailDir` both `EvalSymlinks` the walk start when a ceiling exists. Doing the same here (or comparing with `sameTreeIdentity` instead of `sameCleanPath`) closes it:

```go
ceiling := ""
if top, insideGit := gitWorktreeRootFrom(startDir); insideGit {
	ceiling = top
	if resolved, resolveErr := filepath.EvalSymlinks(startDir); resolveErr == nil {
		startDir = resolved
	}
}
dir := startDir
```

internal/cli/env.go
```diff
@@ -1075,8 +1091,12 @@ func findAmqrcForRoot(root string) (amqrcResult, error) {
 		if absErr != nil {
 			return amqrcResult{}, absErr
 		}
+		ceiling := ""
+		if top, insideGit := gitWorktreeRootFrom(absRoot); insideGit {
+			ceiling = top
+		}
 		dir := absRoot
-		for !isHomeConfigDir(dir) {
+		for ceiling != "" || !isHomeConfigDir(dir) {
```
avivsinai, line 1102: **Same symlink asymmetry as `findRootInParents`, with a second consequence from the loop-condition change.**

The ceiling is computed from the symlink-resolved path (inside `gitWorktreeRootFrom`) while the walk runs on the unresolved `absRoot`, and `sameCleanPath` does not resolve symlinks. `root` here is user-supplied (`--root`, `AM_ROOT`, config values), so symlinked components are entirely realistic. Probe on this branch (nested checkout with its own existing `.agent-mail`, addressed via `parent/alias → parent/nested`):

```text
findAmqrcForRoot via symlinked path adopted parent .amqrc at .../parent/.amqrc
```

Second consequence: the condition changed from `!isHomeConfigDir(dir)` to `ceiling != "" || !isHomeConfigDir(dir)`. When the ceiling is set but never matches (the symlink case), the home-boundary stop is disabled too, so the walk now reads `.amqrc` at every ancestor up to `/` — including `$HOME/.amqrc` mislabeled as a project rc, and e.g. a `/tmp/.amqrc` on the way. That is strictly more permissive than pre-PR behavior in the mismatch case. `findAndLoadAmqrc` (line 758) uses the same loop shape safely only because it resolves its walk start first.

Same fix as `findRootInParents`: `EvalSymlinks(absRoot)` when a ceiling was found (matching `findAndLoadAmqrc` / `detectAgentMailDir`), or compare with `sameTreeIdentity`.

Test gap to cover with the fix, per the repo rule that tests must exercise the actual hostile boundary: add a `TestNestedWorktreeDoesNotAdoptParentLiveBaseRoot` subtest that reaches the nested worktree through a symlinked component (`t.Chdir` sets logical `$PWD`, so `os.Getwd` reproduces it) — the current subtests all use physical paths and pass on a wrong implementation of the ceiling comparison. #663 already treats symlinks as in-scope (`.amqrc` symlink-refuse tests).

avivsinai (commented):
**Verdict: REQUEST_CHANGES** (posted as a comment review — same-account PRs cannot formally request changes)

Reviewed at f2572a1. The direction, scoping, and test approach are right, but the ceiling comparison has a symlink hole that reproduces the exact bug class this PR closes — silent adoption of the parent live queue — so I would block until the two walk starts are symlink-aligned. Details and a verified reproduction are in the line comments; summary below.

### 1. Does the git-worktree ceiling actually isolate a nested/linked worktree?

For physical paths, yes. `gitWorktreeRootFrom` treats any `.git` marker (dir for an independent checkout, file for a linked worktree) as the innermost boundary, fails closed on non-ENOENT lstat errors, and correctly skips a worktree's own `.git` dir in the bare-repo signal. The candidate/`.amqrc` check runs at the ceiling dir before the break, so the worktree root's own config still wins.

**But the ceiling is defeated by a symlinked path component.** The ceiling is computed from the symlink-resolved path while the walk in `findRootInParents` / `findAmqrcForRoot` runs on the unresolved path, and `sameCleanPath` does not resolve symlinks — so the break never fires and the walk crosses into the parent. Verified with a probe on this branch (nested checkout under a parent with `.amqrc` + `.agent-mail`, reached via `parent/alias → parent/nested`):

```text
resolveRoot(".agent-mail") from symlinked cwd adopted parent live queue .../parent/.agent-mail
findRootInParents via symlinked path escaped the ceiling: .../parent/.agent-mail
findAmqrcForRoot via symlinked path adopted parent .amqrc at .../parent/.amqrc
```

This is reachable in production: `os.Getwd` prefers logical `$PWD`, which any `cd` through a symlink produces (aliased checkout dirs, macOS `/tmp`), and `findAmqrcForRoot`'s `root` is user-supplied. The repo already has the exact alignment pattern — `findAndLoadAmqrc` and `detectAgentMailDir` both `EvalSymlinks` the walk start when a ceiling exists — this PR just omitted it at the two new sites. Small, contained fix.

### 2. Fail-closed vs silent parent adoption

The intended behavior is right: no config in the nested worktree → exit 5 context mismatch naming the nested worktree, parent path not leaked, parent tree digest unchanged. The symlink case above is the remaining silent-adoption path. Secondary: in `findAmqrcForRoot`, `ceiling != "" || !isHomeConfigDir(dir)` disables the home-boundary stop whenever the ceiling never matches, so the mismatch case walks past `$HOME` to `/` — strictly more permissive than pre-PR. Both go away with the same one-line alignment.

### 3. Same-worktree scratch subdirs (v0.61 / #663)

Unchanged for physical paths: the walk still checks each level up to and including the worktree root before breaking. Existing `TestSetupProjectRoot*` and discovery tests pass on this branch.

### 4. Tests

Genuine, not a temp-dir rename: real `git init` for the nested independent checkout and real `git worktree add -b linked` for the linked worktree (`.git` file). Assertions are strong — `errAmqrcNotFound`, exit 5 naming the nested worktree and not leaking the parent path, parent tree digest unchanged, `findRootInParents` ceiling, relative `AM_ROOT`, and `setup` writing a local `.amqrc` without touching the parent. All six subtests pass here. Gap: every subtest uses physical paths, so none would catch the `sameCleanPath` hole — add a symlinked-component variant with the fix (`t.Chdir` sets logical `$PWD`, so it reproduces directly).

### 5. Docs / skill / COOP

`COOP.md`, `docs/session-routing.md`, `docs/launch-api.md`, and `skills/amq-cli/SKILL.md` all state the same contract (own config wins, otherwise fail closed, no parent adoption) and match the intended code behavior. Skill version is release-please-managed (`x-release-please-version`), so no manual bump needed; `make check-skills` passes; only the canonical skill dir was edited.

### Test run notes

`go test ./internal/cli` on this VM intermittently fails `TestWakeCleanupDoesNotDeleteReplacement` (wake-lock cleanup, untouched by this PR, disclosed in the PR body as a pre-existing VM failure); it passes on reruns and GitHub CI (`build`, `windows-claim-test`) is green on this PR.

avivsinai (commented):
**Verdict: APPROVE** (follow-up at 5381d96 clears the prior REQUEST_CHANGES)

Reviewed the push that woke this run. `findRootInParents` and `findAmqrcForRoot` now `EvalSymlinks` the walk start when a git ceiling exists — same pattern as `findAndLoadAmqrc` / `detectAgentMailDir` — and the new `symlinked path into nested worktree does not adopt parent` subtest covers the hostile boundary the first review reproduced (logical `$PWD` via alias). Physical-path nested/linked fail-closed behavior and docs/skill wording still look right.

CI on 5381d96 is green (build + windows-claim-test). No remaining marketing/docs concerns for Growth; this is a CLI discovery fix only. Not merging from here.

PR #695, opened 2026-08-30
fix(wake): protect a published successor during cleanup and stop

avivsinai:
## Summary

Wake lifecycle mutation is rewritten behind a single effect boundary instead of being fixed at individual call sites.

An adversarial release review returned NO-GO on the previous approach with six blockers, one of them destructive: guard-absent cleanup could delete a *successor's* published lock, because a retained directory descriptor protects against the directory being replaced but not against a pathname inside that same directory. The other five were the same shape — authority sampled, then an effect or a success report that the sample no longer supported — spread across acquire admission, readiness, Darwin restart, TERM/KILL escalation, and the detached-cleanup error contract.

Six defects of one shape surviving five review rounds is a signal about mechanism, so the fix is structural:

- **`wakeMutationScope`** owns the lifecycle guard and the retained directory descriptor and never exposes the descriptor. Its action methods (`unlinkWakeLock`, `sendPidfdSignal`, `signalProcess`, `queueStopRequest`, `queueRestartSignal`) each revalidate immediately before their effect, so a caller cannot perform one without adjacent authorization.
- **An AST architecture test** fails the build if any file outside the scope files performs a raw wake-lock unlink, a wake process or pidfd signal, or a lifecycle control channel send. It resolves package string constants and **fails closed on any `Unlinkat` whose target it cannot statically resolve**, so indirection through a variable cannot evade it. It reported 19 sites on first run.
- **The directory relation is tri-state with `wakeAgentDirInconclusive` as the zero value**, so a mishandled relation fails closed rather than defaulting to canonical. `ENOENT` on the canonical path is proof of detachment, not uncertainty.
- **Terminate paths acquire the guard with `LOCK_EX|LOCK_NB`.** A wake wedged inside its own guard-held critical section would otherwise hang the CLI at exactly the moment an operator reaches for stop or kill.
- **`assertNotWakeLockName`** is enforced in every cleanup helper whose filename is caller-supplied, so those AST exemptions rest on a local invariant rather than on a claim about current callers.

## Behavior change: Darwin raw numeric termination is `operator_only`

macOS has no pidfd, so nothing pins process identity between inspecting a wake and signalling it — and self-upgrade re-execs in place, keeping the same PID with a new generation, so process identity would not be sufficient even if it were pinned. AMQ therefore refuses PID-based signalling on Darwin rather than guessing. The owning terminal or supervisor stops the wake. This applies to orphan cleanup paths. `amq wake check` reports `restart_capability=operator_only` for these states, so the refusal is predictable before it is encountered.

## Verification

- `go test ./internal/cli`: PASS
- Focused architecture, retire, terminate, and detached-cleanup suites: PASS
- `GOOS=linux` and `GOOS=darwin` compiles: PASS
- `git diff --check`: PASS

TestTerminateWakeHelperProcessRemovesLockCommittedAfterFirstInspection: renamed to TestTerminateWakeHelperProcessPreservesLockCommittedWithoutRetainedDirectory because cleanup must fail closed when it lacks the retained directory authority.
TestDetachedBoundWakeResidueCleanupFailsClosedWhenCanonicalPathAbsent: renamed to TestDetachedBoundWakeResidueCleanupRemovesOldClaimWhenCanonicalPathAbsent because the behavior it pinned was wrong and was deliberately inverted. It previously asserted that an absent canonical path is inconclusive and must preserve the lock; an adversarial review showed that ENOENT is definite proof of detachment rather than uncertainty, so treating it as inconclusive left a real detached lock in place and reported canonical success during the rename-to-absent interval. The renamed test asserts the corrected contract: the exact old claim is removed and `cleanup-only` is returned only after that unlink commits. It asserts strictly more than the original, not less — it additionally pins that canonical-only artifacts, including the wake state file, are left unchanged when detached.
TestPrepareCoopWakeLockLiveRawPartialTakeoverWarnsAndSucceeds: renamed to TestPrepareCoopWakeLockLiveRawPartialTakeoverRefusesAfterLockDisappears to record the semantic inversion: the old test codified the SIGKILL-after-lock-disappearance bug (Pro round-4 B1), while the new test asserts that lock gone plus process still alive refuses SIGKILL with \"wake lock is missing before SIGKILL\".
TestRetireWakeLinuxPreservesRetainedLockWhenCanonicalDirectoryDisappearsAfterStop: renamed to TestRetireWakeLinuxRemovesRetainedLockWhenCanonicalDirectoryDisappearsAfterStop to guard that ENOENT proves detachment, so cleanup removes only the exact retained old claim and leaves canonical-only artifacts unchanged.


avivsinai:
Converting to draft. **Do not merge this branch in its current state.**

An adversarial review gate returned NO-GO with six release blockers, one of them destructive: on the Linux guard-absent cleanup path, the retained directory FD protects against the *directory* being replaced but not against a pathname *inside* it, so a guard-aware successor that publishes a new lock between our final read and the unlink gets **its lock deleted**. That is the exact failure this PR exists to prevent. The remaining five are the same shape — authority sampled, then an effect or a success report that the sample no longer supports — across acquire admission, readiness, Darwin restart, TERM/KILL escalation, and the detached-cleanup error contract.

Six defects of one shape surviving five review rounds is a signal about mechanism, not about diligence. The branch is being reworked from call-site patches to a single `wakeMutationScope` effect boundary: it owns the guard and the retained directory, computes a tri-state canonical relation, exposes only action-specific methods that revalidate immediately before their effect, and is backed by an AST test that rejects raw wake effects outside the scope files — so a future call site cannot reintroduce this class by construction.

The intent of this PR is unchanged and its retire-ordering, schema-remedy, and nlink fail-closed work is retained. It returns to ready when the rewrite passes review and a re-run gate.

PR #696, opened 2026-08-30
fix(wake): bind the inbox parent directory identity through repair

avivsinai:
Binds the inbox parent directory identity through the wake repair handoff, so a replaced parent directory cannot pass the retained-capability check (refs bead agent-message-queue-xxt lineage; originally `fix/wake-repair-component-dirs`, recomposed against current code rather than replayed).

- The parent is obtained via `openat(inboxFD, "..")` rather than a pathname, so a re-parented inbox is caught by identity, not by a string comparison a rename defeats. The identity is carried through the handoff and revalidated in the lineage check.
- Repair handoff schema 1 -> 2. Both rejection sites name a runnable remedy for their own direction: a legacy (schema 1) source says to restart that wake so it re-announces at schema 2; a newer (schema 3) source says to upgrade amq or restart the newer wake under the matching version. A regression asserts both the rejection and the remedy text.
- Mixed-version boundary documented in `docs/wake-lifecycle.md`: a wake started by a pre-schema-2 binary cannot be repaired in place and must be restarted.

**Known limitation, recorded deliberately:** `openat(fd, "..")` carries no blanket cross-filesystem guarantee. Mount and bind-root layouts are unsupported unless the platform returns the exact bound parent; validation fails closed otherwise.

Negative tests include a repaired wake that must NOT inject after inbox-component replacement, symlinked-component rejection, and the Darwin retained-watcher case.

Based on #695 — merge that first. Implemented by codex.

TestTerminateWakeHelperProcessRemovesLockCommittedAfterFirstInspection: renamed to TestTerminateWakeHelperProcessPreservesLockCommittedWithoutRetainedDirectory because cleanup must fail closed when it lacks the retained directory authority.


avivsinai:
Converting to draft. **Do not merge.**

This PR targets `main` and reports MERGEABLE, but it *contains* two commits from #695:

- `8d567d3` fix(wake): protect a published successor during cleanup and stop
- `87c4147` fix(wake): preserve legacy Linux reclaim without guard

#695 was ruled NO-GO by an adversarial gate with six release blockers, and `87c4147` carries the destructive one: on the Linux guard-absent cleanup path a guard-aware successor can have its published lock deleted. Merging this PR as-is would land that defect on `main` through the side door, with #695 correctly held back and looking safe.

The parent-binding work in `50b92dd` is good and is being retained. It needs to be rebuilt on `main` **without** the two inherited commits — cherry-pick `50b92dd` onto a fresh branch off `main`, or rebase this branch once the #695 rewrite lands. Returns to ready when its diff contains only the parent-binding change.

avivsinai:
Superseded by #698. The parent-binding work was rebuilt from main as the clean one-commit branch fix/wake-repair-parent-binding-pr696 at 2eef4d4184f7977e75f1ed087a6826ed4da5ff77, excluding 87c4147 and 8d567d3. The old branch is retained until #698 merges.

PR #697, opened 2026-08-30
feat(keepalive): support native Windows agent adapters

lask3802:
## Summary

Add native Windows implementations for the existing `codex-queue` and
`claude-print` submitted-injection adapters, and publish
`amq-keepalive.exe` as a Windows ZIP.

This deliberately does not advertise `amq wake`, `coop exec`, or terminal
supervision as native Windows features. The Windows support contract is direct
`amq-keepalive.exe inject` into an exact live Codex or Claude Code session.

## Changes

- resolve `codex.exe` / `claude.exe` and accept regular Windows executables
  without Unix mode bits;
- prove Codex writer activity and serialize Claude injects with non-blocking
  `LockFileEx` locks;
- inspect Claude owner PIDs with `OpenProcess` / `WaitForSingleObject`;
- place resumed Claude processes in a Job Object and terminate the complete
  process tree on timeout;
- publish amd64/arm64 Windows keepalive ZIPs and let direct `amq upgrade --all`
  upgrade only the Windows companion that exists;
- add a focused native Windows CI gate and document the narrower capability in
  the install guide, ADR, README, and canonical CLI skill;
- platform-gate three macOS-only cmux assertions and the Unix-only permission
  bit assertion so the complete adapter package can run natively on Windows.

## Testing

- `go test ./internal/keepalive/adapter -count=10` on native Windows;
- hostile parent/grandchild Job Object termination test, 20 runs;
- Windows race detector: full adapter once plus focused Windows/Codex/Claude
  tests for 10 runs (`CGO_ENABLED=1`, MinGW-w64 GCC 16.1.0);
- Gremlins v0.6.0 mutation gate on the new Windows mechanics: 28/28 killed,
  0 lived, 0 uncovered, 100% efficacy and mutant coverage;
- live Codex exact-thread queue test and packaged-binary CLI enqueue;
- live Claude Code 2.1.251 disposable-session resume, including the matching
  post-init `isReplay:true` echo and result;
- GoReleaser v2.18.0 snapshot built all configured targets and produced both
  Windows keepalive ZIPs;
- same-SHA fork GitHub Actions run
  [`33314863690`](https://github.com/lask3802/agent-message-queue/actions/runs/33314863690)
  passed the complete Linux build/test/smoke/lint job, macOS full test, and
  Windows claim plus native Codex/Claude adapter gates;
- clean WSL clone: skill symlink checks, `go vet ./...`, and golangci-lint
  v2.13.1 (`0 issues`) passed. Three WSL full-suite failures were reproduced
  unchanged on pristine v0.75.0 (`253b5ef`).

## Related Issues

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,8 @@ jobs:
         run: go test ./internal/cli -run '^TestDrainInboxItems' -count=1
       - name: Windows launch API compatibility
         run: go test ./launchapi -run '^TestWindowsCompatibilityAdvertisesOnlyCallableLaunchSurface$' -count=1
+      - name: Native Windows Codex and Claude adapters
+        run: go test ./internal/keepalive/adapter -run '^(TestWindows|TestClaudePrint|TestCodexQueue)' -count=1
 
   # Darwin test leg. Linux + a windows claim job run elsewhere; this gates the
   # defect class where a test passes on Linux but fails on macOS (e.g. macOS
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -30,12 +30,13 @@ builds:
       - -s -w -X main.version={{.Version}}
     goos:
       - darwin
+      - windows
     goarch:
       - amd64
       - arm64
     hooks:
       post:
-        - sh -ceu 'codesign --force --sign - --identifier io.github.avivsinai.amq-keepalive "$1"' _ "{{ .Path }}"
+        - sh -ceu 'if [ "$2" = darwin ]; then codesign --force --sign - --identifier io.github.avivsinai.amq-keepalive "$1"; fi' _ "{{ .Path }}" "{{ .Os }}"
 
   - id: amq-bridge
     main: ./cmd/amq-bridge
@@ -84,6 +85,9 @@ archives:
     ids: [amq-keepalive]
     formats: [tar.gz]
     name_template: "amq-keepalive_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
+    format_overrides:
+      - goos: windows
+        formats: [zip]
 
   - id: amq-bridge
     ids: [amq-bridge]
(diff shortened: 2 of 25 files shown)


avivsinai (approved):
Thanks for this — it's careful, well-scoped work, and it landed after two independent reviews (bytes + design).

Highlights from the review:

- **Windows process lifetime is done right.** Wrapping the child in a Job Object and terminating via `TerminateJobObject`, with the child held on stdin until it's assigned to the job, is exactly the correct idiom for a supervised process tree on Windows — and it closes both handles on every error path. The Job handle also pins process identity the way a pidfd does on Linux, which is more than our Darwin path gets.
- **No regression to existing platforms.** The one shared change (`companionBinariesAvailableOn` / `reportUnavailableCompanionsOnWindows`) returns a defensive copy on non-Windows, and the extracted `isExecutableRegularFile` is logically identical to the previous inline check on Unix. macOS/Linux behavior is unchanged.
- **Publishing matches the code.** `.goreleaser.yaml` adds Windows to `amq-keepalive` only; `amq-bridge` and `amq-acp` stay linux+darwin, which is exactly what the runtime companion filter expects — so `--all` on Windows won't chase a binary that isn't published.
- **The writer-lock probe fails closed** and returns a runnable remedy on error, matching the repo's convention.

CI is green (build, macOS, Windows) and the new Windows adapter test guards the behavior. Approving and merging. Thank you for the contribution.

lask3802:
Thanks for the review and approval. The approved upstream CI run is fully green.

One GitHub bookkeeping issue remains: the original empty `ci` and `Gitleaks` check suites are still marked `action_required`, so the PR API reports `mergeStateStatus: UNSTABLE` even though the current commit check rollup is `SUCCESS`. I cannot re-run the Gitleaks suite without repository admin permission; it may need maintainer approval/re-run (or an admin merge).

I have intentionally not pushed a no-op commit, to preserve the reviewed SHA and approval.

PR #731, opened 2026-09-09
test(fsq,remote): committed-durability-uncertain contract regression (B10)

avivsinai:
Pins bead agent-message-queue-611.22.14 (B10) acceptance criteria with regression tests only — the contract already holds on feat/remote-core@0851fe45; these keep it from regressing.

- **fsq**: a directory-sync failure after the tmp -> new rename surfaces as `CommittedDurabilityError` (visible commit, indeterminate durability) whose `FinalPath` names the delivered artifact; the same filename re-delivers to the same path, never a duplicate under a new name. Uses the `SetSyncDirFaultForTest` seam. Deterministic, <1s.
- **remote/core**: a Publisher returning a committed-durability error never advances `PublishedRevision` (the revision was not durably synced), and Reconcile later safely republishes the SAME immutable revision — exactly one extra publish call, no new revision, no duplicate artifact.
- Newly-created ancestor dirs are already parent-synced on this base (0851fe45 fixed the mkdirAllSynced leaf-first walk); the existing layout_sync tests cover it.

Local: full internal/... green, vet clean, golangci-lint 0 issues, fmt-check pass.

Depends on #729 (base branch).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ __pycache__/
 
 # Added by ggshield
 .cache_ggshield
+/amq-remote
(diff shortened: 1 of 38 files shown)


avivsinai:
Closing per review: the committed-durability contract is already guaranteed by 0851fe45 (in #729); these tests pin a Pro-review hypothesis with no observed defect, which fails the no-test-only-PRs bar. The worktrees and local tests remain available if a field defect ever surfaces.

PR #737, opened 2026-09-10
fix(remote/codex): bound writer acquisition + non-lossy reply-required + worker lifecycle (B14b)

avivsinai:
Bead 611.22.15.2 (B14b). Codex transport I/O bounds + callback worker lifecycle, from the B14 decomposition.

- **B9** — separate workerStart/workerStop Onces (the old shared Once meant Close's close(events) never ran → worker leaked every teardown); explicit eventsInFlight counter (waitCallbacks awaits in-flight==0 AND queue empty under the 2s bound); double-Close safe.
- **B6** — wmu is now a cap-1 channel semaphore: acquire selects on ctx.Done() so a wedged writer can't block a control Call past its deadline; deadline install/write/clear under the held slot.
- **B8** — reply-required server requests get their own bounded queue + dedicated worker, non-blocking dispatch; overflow responds an explicit JSON-RPC error (never blocks the read pump, never silently drops). Notifications stay lossy.
- Two extra hardening fixes found during the build: close-frame no longer sent via the unbounded path (a wedged writer could wedge teardown); acceptServerWS builds wsConn via the constructor (a literal left wmu nil after the semaphore switch — nil-chan panic caught by the round-trip test).

Hard-deadline regression tests per blocker (a regression FAILS, never hangs CI). Full internal/remote -race, codex -race x3, vet, golangci-lint, gofmt all green.

### DIFF
--- a/internal/remote/codex/b14b_regression_test.go
+++ b/internal/remote/codex/b14b_regression_test.go
@@ -2,10 +2,12 @@ package codex
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"net"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


avivsinai:
Superseded by the B14b recut on feat/b14b-transport-recut (fixes the Close-teardown send-on-closed panic + removes the dead notification-worker subsystem). New PR supersedes this.

PR #791, opened 2026-09-16
fix(2pt): share raw-status literals between threadStatus and lost-state gate

avivsinai:
Bead agent-message-queue-2pt (P2). The `notLoaded`/`systemError` literals were duplicated between `threadStatus` (Inspect mapping) and the lost-state gate (clearing activeTurn + bumping generation). A new lost-track status would need to be added in both places — easy to miss.

**Failure mode:** adding a new lost-track status in one place but not the other → Inspect shows 'unknown' but the gate doesn't clear activeTurn (or vice versa), causing a stale turn reference.

**Fix:** `isLostStateStatus()` is now the single source of truth. Both `threadStatus` and the lost-state gate call it.

**Verification:** build OK, -race green (11s), vet clean, gofmt clean. RED confirmed: remove `systemError` from helper → Test2ptIsLostStateStatus FAILS.


avivsinai:
Closing as not planned (deferred). Per test-policy review on thread review/batch2-790-797: this is speculative-drift refactor work with no observed defect. threadStatus returns "unknown" on both the helper path and the default path, so the new delegation changes no Inspect semantics; Test2ptThreadStatusUsesSharedHelper checks outputs only and cannot establish delegation. The PR body also advertises a forbidden mutation demonstration ("RED confirmed: remove systemError from helper → Test2ptIsLostStateStatus FAILS"). Repo test rule excludes speculative tests, tests for existing behavior with no observed defect, and mutation demonstrations. A future actual status change can own the minimal classifier refactor and its behavioral test.

PR #823, opened 2026-09-19
fix(config): preserve unknown keys in setup and launch

avivsinai:
Bead c6i — the setup/apply half of review-821-r1 P1-b. Stacked on #821 (feat/611.22.55-config-preserve); base is that branch, not main.

611.22.55 fixed EnsureAgent; two more config.json writers still re-marshaled bare structs and wiped unmodelled keys (default_agent/project/routing) from a live root:

- cli/setup.go marshalSetupBaseConfig now overlays the modelled config onto the raw existing meta/config.json (empty when absent) via the exported config.MarshalPreservingUnknowns; both base and union writes pass the same original bytes.
- launch/apply.go writeApplySessionConfig reads the existing meta/config.json (missing tolerated; other read errors fail closed) and overlays the roster struct the same way.

Byte-stability fix (caught locally): the exported wrapper initially kept a struct-order fast path for empty originals, so a fresh write and its re-read rewrite differed (struct order vs sorted map order) and a matching `amq setup` rerun re-emitted the roster with writes=1, breaking TestSetupWritesAuthoritativeAndPreferenceScopesThenNoOps. MarshalPreservingUnknowns now always overlays (empty original = empty raw map + modelled keys): fresh and rewritten output are byte-identical, unit-pinned.

Tests (red-on-revert verified): TestSetupPreservesUnmodelledConfigKeys, TestWriteApplySessionConfigPreservesUnmodelledKeys, TestMarshalPreservingUnknownsEmptyOriginalIsSortedAndStable. config + launch + cli suites green; vet + golangci-lint clean (only the pre-existing ruled wake_owner_observation_darwin.go fmt drift).


avivsinai:
Recut pushed: a24da8f (feat/611.22.57-config-writers-preserve).

**P2-1 — one encoder for all config.json writers.** `WriteConfig` (amq init / --force), the `ensureAgentLocked` create arm, and `initializeApplySession` (launch apply) now route through `MarshalPreservingUnknowns(nil, …)` like the setup/apply overlay writers already do. Empty original ⇒ struct keys in sorted order, so every writer emits the same layout and `amq setup`'s raw-byte roster comparison sees no phantom change after `amq init`.

**P2-2 — HTML-escape clause.** The unknown-keys preservation doc comment now says explicitly: the compaction pass applies encoding/json's default HTML escaping, so an unknown string value containing `<`, `>`, or `&` changes bytes on the first rewrite (JSON-semantically identical, stable thereafter).

**Regression:** `TestWriteConfigMatchesPreservingLayout` pins WriteConfig's bytes to the encoder's output and asserts sorted key order. RED-verified: reverting WriteConfig to struct-order `json.MarshalIndent` fails the test.

Suites: internal/config + internal/launch green; gofmt/vet/golangci-lint clean; GOOS=windows vet clean.

PR #824, opened 2026-09-19
fix(remote): serve and doctor audit residue (611.13.4)

avivsinai:
Bead 611.13.4 - six one-hunk P2s from review-b7.md on main 3245400 (the #809 post-merge audit); no P0/P1. Tests only where a run showed the defect.

1. **Atomic generated state**: refusals.json and adapters.json now go through fsq.WriteFileAtomic instead of os.WriteFile; a truncated refusals.json used to make doctor silently drop the refusal and exit 0. Doctor now reports `refusals_error` for an unreadable file (not-exist still tolerated).
2. **Attached adapter kind**: attached entries in adapters.json carried an empty kind; the kind now comes from the manifest that published the attachment.
3. **Bounded validation test**: TestValidationFailuresExitTwo drove a real in-process serve with no deadline - under a mutation the package stalled for the full go-test timeout. The serve call is now bounded by a 30s deadline.
4. **ipc.Server.Close ENOENT**: Go unlinks the socket during listener.Close, so the explicit Remove always lost with no-such-file; ENOENT is ignored.
5. **Duplicate-target diagnostic**: serve no longer blames "flags and manifest" when the duplicate was two manifest entries and no flag was passed; the message names the actual sources.
6. **schema_version alignment**: the companion (manifest.Parse) defaults a missing schema_version to 1 while core doctor warned on it; aligned in the companion's direction (missing accepted via *int, explicit incompatible value still warns).

cmd/amq-remote, internal/cli, internal/remote/... green; -race clean on the touched paths.


avivsinai:
Recut pushed: bf86fc8 (feat/611.13.4-serve-doctor-residue).

**Five P1 regression pins, each RED-verified by mutation before landing GREEN:**
1. `refusals_error` — `TestDoctorRefusalsErrorSurfaced`: truncated refusals.json ⇒ key present with the load error; absent file ⇒ key absent. Mutation: dropping the `refusals_error` arm fails the test.
2. Attached-kind duplicate target — `TestStartupSequenceWithManifest` now includes an attached `"fake"` kind case; the refusal must name the target. Mutation (flag-blame wording) fails.
3. Duplicate-target message text pinned to "the same target id was passed more than once" — the old flag-blame wording fails.
4. `TestServerCloseIsIdempotent`: second `Close` returns nil, listener state stays consistent.
5. `TestReadPassiveExtensionManifestAcceptsMissingAndZeroSchemaVersion`: schema_version absent AND explicit 0 both read with zero diagnostics.

**P2s:**
- P2-1 picked a side: the passive reader accepts explicit `schema_version: 0` (the companion defaults 0→1 and cannot distinguish missing from explicit 0; one file must not serve targets=1 and draw a doctor warning simultaneously).
- P2-3 `TestValidationFailuresExitTwo` now shares ONE 30s table-level deadline across all 8 subtests (mutation-verified: a serve regression costs ~30s total, not 8×30s).
- P2-2 doctor wording: `doctor` reporting `endpoint: reachable` with exit 0 is intentional — reachability is the check; session-level refusals surface as `refused session.list:` + exit 6. Documented here rather than changing the exit code (changing it would break the fresh-root path which correctly exits 6 only when the endpoint is absent).

Suites: race -count=3 green on cmd/amq-remote, internal/cli, internal/remote/ipc; gofmt/vet/golangci-lint clean; GOOS=windows vet clean.

PR #887, opened 2026-09-24
feat(skills): /amq-remote connects this session to Buzz in one command

avivsinai:
Bead: agent-message-queue-611.33 (/amq-remote skill).

With this change, inside a running Claude Code or Codex session the user types `/amq-remote`, and the session is connected to their Buzz "AMQ Remote" agent. A DM from the Buzz app or Desktop then runs here. `/amq-remote off` disconnects it, and `/amq-remote status` shows the binding.

- New canonical skill `skills/amq-remote`, symlinked into .claude, .agents and .grok, checked by check-skills, and versioned by release-please.
- The skill runs the one-time steps only when they are missing: the Claude Stop hook (asked once; the doctor names it as `native_capability`) and `amq-acp setup` (one Desktop Import). Then it runs `amq-remote attach --self`.
- Homebrew installs `amq-remote` next to `amq` and `amq-acp`, so `brew install` is the whole install.
- README and INSTALL describe the one-command flow.

Depends on #885 (attach --self, binding mode) and #884 (setup), which land first.

Not live-verified yet. The live check is the owner typing /amq-remote in a fresh session, then DMing from the phone and Desktop.


avivsinai:
Not GO. One finding.

`amq-remote doctor` names `native_capability` only after the endpoint is reachable and `session.list` already contains a `claude_code` target. With no state directory it returns exit 6 on boundary `endpoint` and never reaches that check (`cmd/amq-remote/main.go`, the `os.Stat(stateDir)` early return).

The skill runs doctor before `attach --self`, and its preamble says to stop at the first failure. On a fresh root that stops before the Stop hook and before attach. If the agent ignores the exit and only looks for `native_capability`, that name is absent, so the hook is skipped and attach then admits a Claude target that cannot complete a turn.

The hook step has to happen where doctor can see it: install the Claude Stop hook before relying on that signal, or attach first and install when `failing[].boundary` is `native_capability`. A missing endpoint should not end the connect flow.

The rest is fine: symlinks, check-skills, release-please version line `0.80.1`, and `amq-remote version` matches the Homebrew test (`version` is a real subcommand). CI on `101bf91` is green. Not merged. 611.34 waits until this lands.

